PR #1, opened 2021-05-27
English - Basics/Containers

### LINKED ISSUE #4: English - Basics/Containers

JonathanLorimer:
---

closes #4


soupi:
Hi! Just wanted to mention that since ghc-9.0.1 the default prompt in ghci is simply `ghci>` so you don't have to include `Prelude` and `Data.List` there :)

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
 title: Collections
 ---
 
-Lists, tuples, keyword lists, and maps.
+Tuples, Lists, Assoc Lists, Sets, Maps and Hashmaps.
```
soupi, line 6: It might be useful to mention `Vector` and `Seq` as well?

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
+  :: (Char, Char, Char, Char, Char, Char, Char, [Char])
+```
+__Note__: `:t` in the ghci repl shows the type of the value, it will print out a message of the form `term :: Type`
+
+### Constructing Tuples
+
+We can construct tuples using a familiar syntax `(a,b)`.
+
+```haskell
+Prelude> myTuple = (1 :: Int,"hello")
```
soupi: Can use `True` instead of `1` and you won't have to annotate the type or get a typeclass constraint at the type :)

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
+Prelude> :t myTuple
+myTuple :: (Int, [Char])
+```
+
+### Consuming Tuples
+
+We can consume tuples by breaking them apart. With a 2-tuple the options are pretty simple. They have the predictable names `fst` and `snd`
```
soupi, line 33: maybe mention that these are *functions* which are just available as part of the standard library?

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
+Prelude> fst myTuple
+1
+Prelude> snd myTuple
+"hello"
+```
+
+We can also pattern match on tuples to extract their contents.
+
+```haskell
+Prelude> (\(a,b) -> a + 2) $ myTuple
```
soupi: the `$` is not needed in this case :)

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
+3
+Prelude> (\(a,b) -> b ++ " world") $ myTuple
+"string world"
+```
+
+### Limitations of Tuples
+
+There are a couple of limitations with tuples in haskell. First, they are closed; you cannot add an additional item once the tuple of length n has been defined. Second, the types are fixed; once a tuple has a concrete type you cannot add an element of a different type. You can get around this with specialized functions, but they do not generalize to all tuples lengths.
```
soupi, line 53: Since tuples are immutable, 'adding' a value is not really possible, and 'the types are fixed' are not really meaningful - perhaps it is better to say that it is possible to rephrase that to say something like 'there's no general function that can take a tuple and return a new tuple with one more field'?

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
+Prelude> twoTupleToThreeTuple () myTuple
+(1,"world",())
+```
+
+Here is an example of changing a tuple's type.
+
+```haskell
+Prelude> :t myTuple
+myTuple :: (Int, [Char])
+Prelude> :t substituteFst "hello" myTuple
```
soupi: `substituteFst` is not defined prior to this line, maybe introduce it or alternatively replace it with a lambda?

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
+substituteFst "hello" myTuple :: ([Char], [Char])
+Prelude> substituteFst "hello" myTuple
+("hello","world")
+```
 ## Lists
 
-Lists are simple collections of values which may include multiple types; lists may also include non-unique values:
+From a usability perspective, Lists solve the problem of extensibility that tuples face, but they can only contain one type (in other words they are homogenous). Lists also have a special built in syntax.
```
soupi, line 75: Could be nice to see an erroneous usage of trying to use a list with values of different types and see the error message GHC emits.

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
-```elixir
-iex> [3.14, :pie, "Apple"]
-[3.14, :pie, "Apple"]
+```haskell
+Prelude> [1,2,3,4]
+[1,2,3,4]
 ```
 
-Elixir implements list collections as linked lists.
+Haskell implements list collections as linked lists.
```
soupi, line 82: what is list collections?

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
 This means that accessing the list length is an operation that will run in linear time (`O(n)`).
 For this reason, it is typically faster to prepend than to append:
 
-```elixir
-iex> list = [3.14, :pie, "Apple"]
-[3.14, :pie, "Apple"]
+```Haskell
+iex> list = ["Orange", "Banana", "Apple"]
```
soupi, line 87: `iex` should be replaced with `ghci`

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
+["Orange", "Banana", "Apple"]
 # Prepending (fast)
-iex> ["π" | list]
-["π", 3.14, :pie, "Apple"]
+iex> "Grape" : list
```
soupi, line 90: here as well

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
-[3.14, :pie, "Apple"]
+```haskell
+Prelude> [1,2,3,4]
+[1,2,3,4]
 ```
 
-Elixir implements list collections as linked lists.
+Haskell implements list collections as linked lists.
 This means that accessing the list length is an operation that will run in linear time (`O(n)`).
 For this reason, it is typically faster to prepend than to append:
```
soupi, line 84: Maybe this section will work better after explaining that list is a recursive data type

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
-It is common for tuples to be used as a mechanism to return additional information from functions; the usefulness of this will be more apparent when we get into [pattern matching](../pattern-matching/):
+__Note__: `:{` and `:}` allow you to write multiline definitions in ghci.
 
-```elixir
-iex> File.read("path/to/existing/file")
-{:ok, "... contents ..."}
-iex> File.read("path/to/unknown/file")
-{:error, :enoent}
-```
+Hooray! No more exceptions.
```
soupi, line 221: Worth mentioning that `NonEmpty` also exists

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
+
+```haskell
+instance Semigroup [a] where
+  (<>) = (++)
+```
+
+The keywords aren't important, but it should give you an intuition for when you see `<>` which just means concatenation for a specific type!
 
-### List Subtraction
+### List Difference
```
soupi, line 136: in my opinion this section can be skipped :) users should use a better fitting data type such as `Set` for such use cases anyway.

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
 When using lists, it is common to work with a list's head and tail.
 The head is the list's first element, while the tail is a list containing the remaining elements.
-Elixir provides two helpful functions, `hd` and `tl`, for working with these parts:
 
-```elixir
-iex> hd [3.14, :pie, "Apple"]
-3.14
-iex> tl [3.14, :pie, "Apple"]
-[:pie, "Apple"]
+Haskell provides two helpful functions, `head` and `tail`, for working with these parts:
```
soupi, line 169: might be worth including simple pattern matching as well.

soupi (commented):
I like this overall and have left a few comments, also another thing that I think is the most important is mentioning *when* each data structure is a good choice.

en/lessons/basics/collections.md
JonathanLorimer: I think pattern matching is covered in basics and I roughly demonstrate it in the `safeHead` example. I am going to leave this comment unresolved though, I think its a worthwhile subject if pattern matching isn't covered elsewhere.

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
 ---
 
-Lists, tuples, keyword lists, and maps.
+Tuples, Lists, Assoc Lists, Sets, Maps/Hashmaps, Seqs, and Vectors.
 
 {% include toc.html %}
 
+## Tuples
+
+Tuples are the first kind of collection you are introduced to in haskell. This is because they are simple, primitive, and have a terse, flexible, built in syntax. Tuples are sometimes referred to as anonymous records, but instead of referencing fields by name you reference them by position in the structure. Tuples can contain an arbitrary number of items, but we will focus on the 2-tuple since haskell has good built in support for them. However, if you are interested this is what an 8-tuple looks like
```
Bodigrim, line 12: 1. Can we wrap lines at 80 symbols? It would allow to pinpoint comments and propose suggestions more precisely.
2. Tuples cannot contain an arbitrary number of items, only up to 15-tuple is [required](https://www.haskell.org/onlinereport/basic.html#basic-tuples) by Haskell Report. GHC supports up to 62-tuple.
3. "haskell" -> "Haskell" here and below.

en/lessons/basics/collections.md
JonathanLorimer: Interesting, I kind of agree, but left it to keep semi-parity with the elixir example. Would be interested to hear other's thoughts on it. I also thought about going the other way and showing all the set operations on lists, but that seemed like unnecessary complexity.

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
+
+```haskell
+ghci> :t ('0','1','2','3','4','5','6',"8-tuple")
+('0','1','2','3','4','5','6',"8-tuple")
+  :: (Char, Char, Char, Char, Char, Char, Char, [Char])
```
Bodigrim, line 17: Type declarations have spaces after commas, but terms do not. It would be nice to make it consistent (I suggest putting spaces in terms as well).

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
+ghci> twoTupleToThreeTuple () myTuple
+(1,"world",())
+```
+
+Here is an example of changing a tuple's type.
+
+```haskell
+ghci> :t myTuple
+myTuple :: (Int, [Char])
+ghci> :t substituteFst "hello" myTuple
```
Bodigrim, line 68: Where is `substituteFst` defined?

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,224 @@ version: 1.3.1
...
+
+```haskell
+ghci> head ["Orange", "Banana", "Apple"]
+"Orange"
+ghci> tail ["Orange", "Banana", "Apple"]
+["Banana","Apple"]
 ```
 
-In addition to the aforementioned functions, you can use [pattern matching](../pattern-matching/) and the cons operator `|` to split a list into head and tail. We'll learn more about this pattern in later lessons:
+Unfortunately these functions reveal an ugly part of the language's base library, some of the functions are partial. This means that they do not cover the full domain of possible inputs.
```
Bodigrim, line 178: It would be nice to give an example here.

en/lessons/basics/collections.md
soupi: Yeah it makes a lot of sense to keep semi-parity with the elixir guides. In this case though I'd say that it isn't a good idea -  because lists are such a convenient data structure we unfortunately tend to use them in places where they are inappropriate, so at least I suggest not demonstrating "bad practices" (those that will lead to incorrect results or bad performance) in tutorials even if Elixir do it :slightly_smiling_face:
And on the plus side it's less work :smiley:

soupi: Btw this is why I would also suggest to leave out association lists in favor of `Map` :)

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,287 @@ version: 1.3.1
...
+        Actual: Bool
+    ...
+```
+
+### Inductive Types
+
+Lists are also a first introduction to an "inductive" or "recursive" type. Here is an example that is identical to the haskell implementation without the syntactic sugar (we use backticks to add infix syntax).
+
+```haskell
+data List a = Nil | a `Cons` List a
```
soupi, line 118: Do you think that maybe this should be `Cons a (List a)` instead?

JonathanLorimer: Yeah, I was concerned about getting too deep into fixity here. I just wanted to keep the syntax similar to the actual list syntax.

en/lessons/basics/collections.md
```diff
@@ -3,120 +3,287 @@ version: 1.3.1
...
-```elixir
-iex> {3.14, :pie, "Apple"}
-{3.14, :pie, "Apple"}
+__Note__: `:i` in ghci will give you some information about the type, the first line is the implementation.
+
+Now we can define a total head and tail function using pattern matching!
+
+```haskell
+ghci> :{
+ghci Data.List| safeHead :: [a] -> Maybe a
```
soupi, line 236: Can remove the `Data.List` from here as well :)

en/lessons/basics/collections.md
JonathanLorimer: Okay, you've convinced me!

en/lessons/basics/collections.md
ulysses4ever: I agree: that infix and expecialy ticks may be more confusing. My vote would be to hold off such tricks till the chapter on data types.

en/lessons/basics/collections.md
JonathanLorimer: whoops, the example below was supposed to be just this!

en/lessons/basics/collections.md
```diff
@@ -3,138 +3,363 @@ version: 1.3.1
...
+## Tuples
 
-Lists are simple collections of values which may include multiple types; lists may also include non-unique values:
+Tuples are the first kind of collection you are introduced to in Haskell. This
+is because they are simple, primitive, and have a terse, flexible, built in
+syntax. Tuples are sometimes referred to as anonymous records, but instead of
+referencing fields by name you reference them by position in the structure.
+Tuples can contain an arbitrary number of items, well this is not entirely
+true, the Haskell Report only requires up to 15-tuples and GHC comes with
+support for 62-tuples, but usually that is enough.We will focus on the 2-tuple
```
myShoggoth, line 18: Needs a space before "We"

en/lessons/basics/collections.md
```diff
@@ -3,138 +3,363 @@ version: 1.3.1
...
-Lists, tuples, keyword lists, and maps.
+Tuples, Lists, Assoc Lists, Sets, Maps/Hashmaps, Seqs, and Vectors.
 
 {% include toc.html %}
 
-## Lists
+## Tuples
 
-Lists are simple collections of values which may include multiple types; lists may also include non-unique values:
+Tuples are the first kind of collection you are introduced to in Haskell. This
```
myShoggoth, line 12: "kind of" is extra

en/lessons/basics/collections.md
```diff
@@ -3,138 +3,363 @@ version: 1.3.1
...
+is because they are simple, primitive, and have a terse, flexible, built in
```
myShoggoth, line 13: Can cut "This is because"

en/lessons/basics/collections.md
```diff
@@ -3,138 +3,363 @@ version: 1.3.1
...
+syntax. Tuples are sometimes referred to as anonymous records, but instead of
+referencing fields by name you reference them by position in the structure.
+Tuples can contain an arbitrary number of items, well this is not entirely
```
myShoggoth, line 16: A little awkward, but you can retain the same feel with "Although tuples can theoretically contain an arbitrary number of items, the reality is..".  If I were feeling witty I might say "the grim reality is".

en/lessons/basics/collections.md
```diff
@@ -3,138 +3,363 @@ version: 1.3.1
...
+ghci> :t curry (uncurry (&&))
+curry (uncurry (&&)) :: Bool -> Bool -> Bool
+```
 
-```elixir
-iex> [3.14, :pie, "Apple"]
-[3.14, :pie, "Apple"]
+### Constructing Tuples
+
+We can construct tuples using a familiar syntax `(a,b)`.
```
myShoggoth, line 55: If they're new to Haskell I'm not sure if it would be familiar.

en/lessons/basics/collections.md
JonathanLorimer: Andrew, you cheeky devil. "grim reality" it is!

en/lessons/basics/collections.md
```diff
@@ -3,138 +3,363 @@ version: 1.3.1
...
+
+```haskell
+ghci> fst myTuple
+True
+ghci> snd myTuple
+"hello"
 ```
 
-### List Concatenation
+The types of these functions simple, yet reveal exactly how they behave. `fst`
```
myShoggoth, line 76: are simple

en/lessons/basics/collections.md
```diff
@@ -3,138 +3,363 @@ version: 1.3.1
...
 
-+ Keys are atoms.
-+ Keys are ordered.
-+ Keys do not have to be unique.
+It is also interesting to note the `Eq a` constraint on the "key" which allows
+the lookup function to do an equality comparison on them.
 
-For these reasons, keyword lists are most commonly used to pass options to functions.
+While assoc lists are a nice introduction to key value collections that build
+on the previous types we learned about they are not particularly useful. A list
```
myShoggoth, line 357: comma after "about"

en/lessons/basics/collections.md
ulysses4ever: @myShogoth for trivial changes it may make sense to use GitHub's "suggestions" mechanism: just as much typing for you, and one-click approval for the author...

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,757 @@ version: 1.3.1
...
+lookup asymptotics (`O(log n)`) due to the ordering on the key values. The
+example that immediately springs to mind is session storage on the server, in a
+web application. The session state can be indexed by an id that is stored in
+the cookie.
+
+## Hashmaps
+
+There are some cases where we do not have an ordering on our type, but still
+want to use it as a key to index a map. In this case we probably want to reach
+for a hashmap. A hashmap simply hashes the key, et voilà, we have an ordering!
```
JonathanLorimer, line 668: No need to translate this one to french! I got it!

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
+
+While assoc lists are a nice introduction to key value collections that build
+on the previous types we learned about, they are not particularly useful. A list
+simply isn't a very good data structure for lookup, as it provides worst case
+`O(n)` asymptotics. Assoc lists are usually an intermediate data structure
+which Haskell programmers will usually convert into a `Map`. Although this
+conversion is itself an `O(n*log n)` operation `Map` provides an `O(log n)`
+lookup, so the conversion cost is ammortized very quickly.
+
+## Sets
```
Kleidukos, line 366: This part is very complete but I fear that we are putting too much out there!  
Could you please take two, three sentences for each "major" operation? I think the bulk of the content of this section is more appropriate for a post in the "blog" section of the website (where we can be much more prolific)

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
+
+There are some cases where we do not have an ordering on our type, but still
+want to use it as a key to index a map. In this case we probably want to reach
+for a hashmap. A hashmap simply hashes the key, et voilà, we have an ordering!
+This does require that the key type is hashable though.
+
+The module that exports the `HashMap` data type and functionality is called
+`Data.HashMap.Strict`, it lives in the `unordered-containers` package.
+Unfortunately `unordered-containers` are not included in ghci (as of 9.0.1) so
+you will have to setup a project in order to pull in that dependency. Fret not
```
Kleidukos, line 674: You can remove this section, using external packages on-the-fly is covered by @soupi [here](https://github.com/haskell-infra/www.haskell.org/pull/88/files#diff-6e06515857e7603e2d734c3a97cde77cf7d00344596c99c8d257e98e5526b49aR236), and I intend to link to this page at the beginning of the Basics track.

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
+indexed access. The `vector` library offers both a safe and unsafe index
+accessor operator.
+
+Here is an example where we want to create a container that allows us to
+retrieve ascii chars by their character code.
+
+__Note__: Because this library is not a core library it doesn't ship with ghci,
+therefore you will not be able to run the code below. However, if you have
+`cabal-install` available you can run `cabal new-repl --build-depends "vector"`
+and then the snippet below should work.
```
Kleidukos, line 696: Same as above, no need to make a fuss out of it. Just say something along the lines of "you can fetch this package by typing --build-depends when invoking `cabal repl`" (and no need for `new-`, with Cabal 3.4 every command is `v2-` or `new-`)

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
+'A'
+ghci> asciiChars !? 65
+Just 'A'
+ghci> asciiChars !? 128
+Nothing
+ghci> asciiChars !? 127
+Just '\DEL'
 ```
+
+Vectors also offer efficient slicing `O(1)`, so they are really good for
```
Kleidukos, line 715: nitpick: You can use `𝛰` instead of `O` when it comes to Omicron complexity

Kleidukos (changes requested):
Here is a very big request for changes! I highlighted only the points to change, but I want to make sure you know that I appreciate you work. :) Thanks a lot!

en/lessons/basics/collections.md
Kleidukos: I'm not sure about Seq. From what I hear and see (and use), it's more or less at the same level as  `Array i e ` nowadays…

soupi (changes requested):
Oh my god I didn't click the submit review..... submitting.

Very good work btw!

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
+"Orange"
+ghci> tail ["Orange", "Banana", "Apple"]
+["Banana","Apple"]
 ```
 
-Be mindful of duplicate values.
-For every element on the right, the first occurrence of it gets removed from the left:
+Unfortunately these functions reveal an ugly part of the language's base
+library, some of the functions are partial. This means that they do not cover
+the full domain of possible inputs.
```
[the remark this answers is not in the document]
soupi: Maybe combine the two? "may raise an exception, even when given an argument with an appropriate type, because they do not cover the full domain of possible inputs" this way we include both the cause and the effect

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
+we have a lot of information about the structure of our data and therefore
+access is very perfomant `O(1)`. From a type safety perspective tuples are
+useful for associating data; saying "I required both of these things". For
+example if we wanted to represent a money transfer it wouldn't make sense to
+have just the amount or just the destination, we would want to group them `(Int,
+AccountId)`.
+
+There is an interesting parallel between functions that require n arguments and
+functions that require n-tuples. While "curried" functions are arguably more
+useful due to partial application, these two modalities are equivalent.
```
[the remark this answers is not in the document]
ulysses4ever: I think this is a divergence here and the discussion belongs elsewhere.

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
+myTuple :: (Bool, [Char])
+```
+
+### Consuming Tuples
+
+We can consume tuples by breaking them apart. With a 2-tuple the options are
+pretty simple. They have the predictable names `fst` and `snd`, and are
+available in the standard library, so no need to import anything! 2-tuples are the only
+arity of tuple that comes with this kind of built in support, so if you need anything larger
+be prepared to write your own accessor functions.
```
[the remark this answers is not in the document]
JonathanLorimer: I do mention pattern matching below this, do you want me to cover it in more depth?

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
+    • Couldn't match expected type: (a, b)
+                  with actual type: (a0, b0, c0)
+                  ...
+```
 ## Lists
 
-Lists are simple collections of values which may include multiple types; lists may also include non-unique values:
+From a usability perspective, Lists solve the problem of extensibility that
+tuples face, but they can only contain one type (in other words they are
+homogenous). Lists also have a special built in syntax.
```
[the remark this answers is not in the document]
JonathanLorimer: nice catch!

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
-```elixir
-iex> [head | tail] = [3.14, :pie, "Apple"]
-[3.14, :pie, "Apple"]
-iex> head
-3.14
-iex> tail
-[:pie, "Apple"]
+Notice that this time the error is not a runtime exception but a type error, the
+compiler is politely telling us that we tried to use a (potentially empty) list
+rather than the required `NonEmpty` type.
```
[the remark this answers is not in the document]
ulysses4ever: I wonder if "politely" here is really necessary. For me it's a distraction. But maybe someone likes this kind of thing.

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
 
-A side note about the name (`++/2`) format used above:
-In Elixir (and Erlang, upon which Elixir is built), a function or operator name has two components: the name you give it (here `++`) and its _arity_.
-Arity is a core part of speaking about Elixir (and Erlang) code.
-It is the number of arguments a given function takes (two, in this case).
-Arity and the given name are combined with a slash. We'll talk more about this later; this knowledge will help you understand the notation for now.
+A side note about another operator that you might come across `<>`. This
+operator is sometimes referred to as `mappend`, it represents generic
+concatenation. In other words any type that supports `<>` can be "appended". In
+the case of lists `<>` is exactly `++` under the hood.
```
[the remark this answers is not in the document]
JonathanLorimer: Like at the beginning of the list lesson, or at the beginning of the containers section? I already received feedback to make this explanation more simple, and I think having it located next to list concatenation gives important context. But I could be convinced otherwise!

en/lessons/basics/collections.md
JonathanLorimer: I ended up not covering seq. I don't personally use it very often, and don't understand it very well. If someone wants to add it in a subsequent PR I think that is fine.

en/lessons/basics/collections.md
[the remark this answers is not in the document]
JonathanLorimer: I struggled a bit to find the correct tone for this medium. I think a bit of levity can be nice for beginners, but I didn't want it to be distracting or verbose. I think I agree politely can be removed here, erring on the side of restraint is probably a good policy.

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
+type NonEmpty :: * -> *
+data NonEmpty a = a :| [a]
+  	-- Defined in ‘GHC.Base’
 ```
 
-In addition to the aforementioned functions, you can use [pattern matching](../pattern-matching/) and the cons operator `|` to split a list into head and tail. We'll learn more about this pattern in later lessons:
+From its definition we can see that `NonEmpty` requires the first element to be
+present. Despite its similarity to the sum constructor `|`, the `:|` constructor
+of `NonEmpty` is just a plane old binary constructor; just a fancy looking
+tuple!
```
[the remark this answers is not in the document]
JonathanLorimer: Tell me if you like this new sentence better.

en/lessons/basics/collections.md
JonathanLorimer: You're breaking my heart! But I think I agree, I will try and save the content for a blog post, and shorten this section.

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,776 @@
...
+  :: (Char, Char, Char, Char, Char, Char, Char, [Char])
+```
+__Note__: `:t` in the ghci repl shows the type of the value, it will print out a
+message of the form `term :: Type`
+
+### When to Use
+
+Tuples are not very flexible because they have a fixed length. This also means
+we have a lot of information about the structure of our data and therefore
+access is very perfomant `O(1)`. From a type safety perspective tuples are
```
TrueBoxGuy, line 33:
```suggestion
access is very performant `O(1)`. From a type safety perspective tuples are
```

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,776 @@
...
+(&&) :: Bool -> Bool -> Bool
+ghci> :t uncurry (&&)
+uncurry (&&) :: (Bool, Bool) -> Bool
+ghci> :t curry (uncurry (&&))
+curry (uncurry (&&)) :: Bool -> Bool -> Bool
+```
+
+### Constructing Tuples
+
+We can construct tuples using parnetheses with the elements separated by a
```
TrueBoxGuy, line 54:
```suggestion
We can construct tuples using parentheses with the elements separated by a
```

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,776 @@
...
+It is also interesting to note the `Eq a` constraint on the "key" which allows
+the lookup function to do an equality comparison on them.
+
+While assoc lists are a nice introduction to key value collections that build
+on the previous types we learned about, they are not particularly useful. A list
+simply isn't a very good data structure for lookup, as it provides worst case
+`O(n)` asymptotics. Assoc lists are usually an intermediate data structure
+which Haskell programmers will usually convert into a `Map`. Although this
+conversion is itself an `O(n*log n)` operation `Map` provides an `O(log n)`
+lookup, so the conversion cost is ammortized very quickly.
```
TrueBoxGuy, line 364:
```suggestion
lookup, so the conversion cost is amortised very quickly.
```
This uses the recommended British spelling.

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,776 @@
...
+ghci> insert1 = insert 1
+ghci> insert1 empty
+fromList [1]
+ghci> insert1 (insert1 empty)
+fromList [1]
+ghci> insert1 (insert1 (insert1 empty))
+fromList [1]
+```
+__Note__: Idempotence is a property the some functions have. The property
+characterized by a function the yields the same result no matter how many times
```
TrueBoxGuy, line 420:
```suggestion
characterised by a function the yields the same result no matter how many times
```
This uses the recommended British spelling.

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,776 @@
...
+Or by creating a set from a list
+
+```haskell
+ghci> fromList [4,3,2,1]
+fromList [1,2,3,4]
+```
+
+You might notice that the elements are sorted after turning them into a `Set`.
+Internally the `Set` data type depends on its contents being orderable. The
+concrete implementaiton of sets in haskell are binary trees, which depend on an
```
TrueBoxGuy, line 394:
```suggestion
concrete implementation of sets in Haskell are binary trees, which depend on an
```

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,776 @@
...
+False
+```
+
+This function, like `difference`, has an order to the arguments. We are asking
+if the first argument is a subset of the second. From the example above you can
+see that `vowels` are a subset of the `alphabet` but not the other way around.
+
+### Cartesian Products
+
+Cartesion products, sometimes referred to as cross multiplication, allow us to
```
TrueBoxGuy, line 559:
```suggestion
Cartesian products, sometimes referred to as cross multiplication, allow us to
```

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,776 @@
...
+True
+ghci> isEven n = member n evens
+ghci> isEven 7
+False
+ghci> isEven 8
+True
+```
+
+You might say "hmmm a `1000000` limit for even numbers seems incorrect",
+and you would be right! This highlights a property of sets in haskell, they
```
TrueBoxGuy, line 442:
```suggestion
and you would be right! This highlights a property of sets in Haskell, they
```

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,776 @@
...
+technical interview.
+
+## Maps
+
+In Haskell, maps are the "go-to" key-value store, sometimes referred to as a
+dictionary.
+
+A `Map` requires that its key has an ordering (`Ord k`), in the same way that
+`Set` required an ordering on the value. This is because the internal
+implementation of the `Map` type in haskell is a size balanced binary tree. This
```
TrueBoxGuy, line 591:
```suggestion
implementation of the `Map` type in Haskell is a size balanced binary tree. This
```

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,776 @@
...
+fromList [("first",3)]
+ghci> :t adjust
+adjust :: Ord k => (a -> a) -> k -> Map k a -> Map k a
+ghci> adjust (+2) "second" oneItem
+fromList [("first",1)]
+```
+
+The observant reader will notice that `adjust` doesn't actually update the map.
+The second invocation of adjust returns the original `oneItem` map, this makes
+sense when you consider that all data in haskell is immutable!
```
TrueBoxGuy, line 651:
```suggestion
sense when you consider that all data in Haskell is immutable!
```

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
+ghci> :t ('0', '1', '2', '3', '4', '5', '6', "8-tuple")
+('0', '1', '2', '3', '4', '5', '6', "8-tuple")
+  :: (Char, Char, Char, Char, Char, Char, Char, [Char])
+```
+__Note__: `:t` in the ghci repl shows the type of the value, it will print out a
+message of the form `term :: Type`
+
+### When to Use
+
+Tuples are not very flexible because they have a fixed length. This also means
```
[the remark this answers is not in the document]
JonathanLorimer: Yeah, I think this makes a lot of sense

en/lessons/basics/collections.md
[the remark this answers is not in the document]
JonathanLorimer: agreed.

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
+
+```haskell
+ghci> myTuple = (True,"hello")
+ghci> :t myTuple
+myTuple :: (Bool, [Char])
+```
+
+### Consuming Tuples
+
+We can consume tuples by breaking them apart. With a 2-tuple the options are
```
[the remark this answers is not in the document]
JonathanLorimer: Good point!

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
+ghci> 1 : 2 : 3 : 4 : []
+[1,2,3,4]
+ghci> Cons 1 (Cons 2 (Cons 3 (Cons 4 Nil)))
+Cons 1 (Cons 2 (Cons 3 (Cons 4 Nil)))
+```
+
+### List Performance
+
+Haskell implements lists as linked lists. Where the cons cells (this operator is
+called a cons `:`) act as the links.  This means that accessing the list length
```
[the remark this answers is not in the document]
JonathanLorimer: I like your explanation, I am going to use it. I also love your quick heuristic for which data structure to choose, I think that makes sense to add as a tl;dr at the beginning.

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
+ghci> import Data.List.NonEmpty
+ghci> :i NonEmpty
+type NonEmpty :: * -> *
+data NonEmpty a = a :| [a]
+  	-- Defined in ‘GHC.Base’
 ```
 
-In addition to the aforementioned functions, you can use [pattern matching](../pattern-matching/) and the cons operator `|` to split a list into head and tail. We'll learn more about this pattern in later lessons:
+From its definition we can see that `NonEmpty` requires the first element to be
+present. Despite its similarity to the sum constructor `|`, the `:|` constructor
```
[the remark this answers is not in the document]
JonathanLorimer: ahhh, very good point.

en/lessons/basics/collections.md
```diff
@@ -3,209 +3,774 @@ version: 1.3.1
...
+ghci> strToChar = [("a", 'a'), ("b", 'b'), ("c", 'c')]
+ghci> fromList strToChar
+fromList [("a",'a'),("b",'b'),("c",'c')]
+ghci> strToCharMap = fromList strToChar
+ghci> :t strToCharMap
+strToCharMap :: Map String Char
+ghci> Data.Map.lookup "a" strToCharMap
+Just 'a'
+```
+__Note__: You might remember that there is a lookup function that works on
```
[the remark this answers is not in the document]
JonathanLorimer: Good point!

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,799 @@
...
+
+## Updating Values to a Map
+
+A useful function to be aware of is `adjust`. This lets us update a value at a
+specified key, only if it exists, if not the old map is returned.
+
+```haskell
+ghci> Map.adjust (+2) "first" oneItem
+fromList [("first",3)]
+ghci> :t Mapadjust
```
TrueBoxGuy, line 674:
```suggestion
ghci> :t Map.adjust
```

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,799 @@
...
+only requires up to 15-tuples and GHC comes with support for 62-tuples. We will
+focus on the 2-tuple since Haskell has good built in support for them. Just so
+you can see, here is an example of a tuple with 8 fields.
+
+```haskell
+ghci> :t ('0', '1', '2', '3', '4', '5', '6', "8-tuple")
+('0', '1', '2', '3', '4', '5', '6', "8-tuple")
+  :: (Char, Char, Char, Char, Char, Char, Char, [Char])
+```
+__Note__: `:t` in the ghci repl shows the type of the value, it will print out a
```
TrueBoxGuy, line 90: There is a comma splice here.

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,799 @@
...
+
+Maps are really great for in memory persistence of state that will need to be
+retrieved by a key of some arbitrary type. This is is because maps have great
+lookup asymptotics (`𝛰(log n)`) due to the ordering on the key values. The
+example that immediately springs to mind is session storage on the server in a
+web application. The session state can be indexed by an id that is stored in
+the cookie, and the session state can be retrieved from an in memory `Map` on
+every request. This solution doesn't scale infinitely, but you would be
+surprised how well it works!
+
```
ulysses4ever, line 694: I’m not sure HashMaps are even worth mentioning in this supposedly introductory text. [There’s a recent post](https://oleg.fi/gists/posts/2021-05-19-dont-default-to-hashmap.html) by @phadej arguing that HashMaps are overrated...

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,799 @@
...
+ghci> isEven 8
+True
+```
+
+You might say "hmmm a `1000000` limit for even numbers seems incorrect", and you
+would be right! This highlights a property of sets in haskell, they are finite
+due to the strictness in the underlying data structure. This contrasts lists,
+which are lazy and therefore potentially infinite.
+
+### Difference
```
ulysses4ever, line 520: This series of set-related operations take incredibly large space for such a niche instrument: I rarely find myself taking set differences, for example. I think 20% of interface (really: insert and lookup) cover 80% of users.

More concretely, I’d suggest to pack the set-theoretic operations (diff, union, intersect and subset; maybe Cartesian product) into one section with a single block of examples. This is a drastic change, so maybe if others are really happy with the current version, I don’t want to insist. But I wholeheartedly believe this topic is currently vastly overweighted with little benefit.

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,799 @@
...
+Similarly, in order to append a list to an existing list, we need to go to the
+end of the existing list, and add a cons cell that points to the new list:
+
+```haskell
+append originalList newList =
+    case originalList of
+        [] -> newList
+        x : xs -> x : append xs newList
+```
+
```
ulysses4ever, line 404: There should be a mention of `++` in particular, I think: its usage really is a code smell for anything more complex than hello ++ world. In particular, I’d mention as an example that having ++ in a loop brings you quadratic performance.

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,799 @@
...
+[1,2,3,4]
+ghci> ["hello", True]
+
+<interactive>:22:11: error:
+    • Couldn't match type ‘Bool’ with ‘[Char]’
+      Expected: String
+        Actual: Bool
+    ...
+```
+
```
ulysses4ever, line 209: I’m still confused  as to why would you put it here. There will be a chapter on user-defined datatypes, but I don’t think you need this intro here. And I don’t think it helps to understand what follows in any way.

ulysses4ever (commented):
Some more comments, mostly about stuff that I think should be removed as not really belonging here. Thanks a lot for your work!

en/lessons/basics/containers.md
JonathanLorimer: hmmmm, I would be interested in @Kleidukos thoughts here. Its a really small section and doesn't add that much, so I could see how it would be more confusing than beneficial.

en/lessons/basics/containers.md
JonathanLorimer: Would be interested to hear what others think

en/lessons/basics/containers.md
JonathanLorimer: In another comment someone asked to see the error of having multiple types in a list. I think we are getting to a point where different opinions are at play. Would be interested what other people think about this, and I am happy to remove if that's the consensus.

TrueBoxGuy: I feel that, for beginners, we should try to give a terse overview of the language that gives them the impression that they could use the language themselves. Showing too many type errors may be off-putting as it slows down the pace of progress they feel they make and their sense of understanding of the language.

en/lessons/basics/containers.md
TrueBoxGuy: Shortening the page (by grouping operations together and using more terse language) would likely make it more digestible for readers and increase the speed of the progress they make.

en/lessons/basics/containers.md
Kleidukos: HashMap is still a pretty big part of this ecosystem, despite its warts. Let us keep this section for now.

en/lessons/basics/containers.md
Kleidukos: As said in my other review, I would be absolutely delighted if the Set part could be a fully-fledged blog post on Haskell School. The point of these lessons is to give a touch and feel of the language, sometimes at the price of exhaustiveness.

en/lessons/basics/containers.md
Kleidukos: I agree with @TrueBoxGuy on this one, the pedagogical mechanics of "Do this in order to have that" (what we do) and "Here is how it can go wrong" are quite different (and the latter would be more suited for a bigger format, like a dedicated blog post)

ulysses4ever: Friends, I'm sorry: my comment was meant for the next section where a manual definition of list is given. I don't mind the examples of errors.

en/lessons/basics/containers.md
JonathanLorimer: It seems there is consensus!

en/lessons/basics/containers.md
JonathanLorimer: Ahhh, I was trying to demonstrate how the list data type works under the hood (since the built in syntax kind of hides that). I thought lists were a good moment to introduce inductive / recursive types. If you think that is valuable, but wasn't clear, I can try and make it more explicit. If you think that this topic is out of place and distracting, I am happy to move it.

ulysses4ever: Yeah, i was arguing for the latter, that is: not having it in this document at all because user-defined types is a huge topic on itself. Any online tutorial on basic data structures I can remember  (tuples/lists first and foremost, and not only in Haskell, other languages too) shoes how to use lists but not how to define them. Understanding of data-definitions shouldn't be required to understand how to use lists in Haskell! Data-definitions should be the next step instead.

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,752 @@
...
+Hooray! No more exceptions.
+
+Another container that is the cousin of the list, and ensures that `head` and
+`tail` are safe is the nonempty list.
+
+```haskell
+ghci> :i NE.NonEmpty
+type NE.NonEmpty :: * -> *
+data NE.NonEmpty a = a NE.:| [a]
+  	-- Defined in ‘GHC.Base’
```
Kleidukos, line 336:
```suggestion
Another way to ensures that `head` and `tail` are safe is the non-empty list:

```haskell
ghci> import Data.List.NonEmpty
ghci> :i NonEmpty
data NonEmpty a = a .:| [a]
```

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,752 @@
...
+ghci> :t NE.head
+NE.head :: NE.NonEmpty a -> a
+ghci> NE.head (1 NE.:| [])
+1
+ghci> NE.head []
+
+<interactive>:11:9: error:
+    • Couldn't match expected type: NE.NonEmpty a
+                  with actual type: [a0]
+    ...
```
Kleidukos, line 358:
```suggestion
```haskell
ghci> import qualified Data.List.NonEmpty as NE
ghci> :t NE.head
NE.head :: NonEmpty a -> a
ghci> head (1 :| [])
1
ghci> NE.head []

<interactive>:11:9: error:
    • Couldn't match expected type: NonEmpty a
                  with actual type: [a]
    ...
```

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,752 @@
...
+
+It is also interesting to note the `Eq a` constraint on the "key" which allows
+the lookup function to do an equality comparison on them.
+
+While assoc lists are a nice introduction to key value containers, and they build
+on the previous types we learned about, they are not particularly useful. A list
+simply isn't a very good data structure for lookup, as it provides worst case
+`𝛰(n)` asymptotics. Assoc lists are usually an intermediate data structure
+which Haskell programmers will convert into a `Map`. Although this
+conversion is itself an `𝛰(n*log n)` operation `Map` provides an `O(log n)`
```
Kleidukos, line 443:
```suggestion
conversion is itself an `𝛰(n*log n)` operation `Map` provides an `𝛰(log n)`
```

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,752 @@
...
+A set can be constructed exclusively by inserting elements into the empty set
+
+```haskell
+ghci> :t Set.empty
+Set.empty :: Set.Set a
+ghci> Set.empty
+fromList []
+ghci> Set.insert 1 (Set.insert 2 (Set.insert 3 Set.empty))
+fromList [1,2,3]
+```
```
Kleidukos, line 461:
```suggestion
```haskell
ghci> import Data.Set (Set)
ghci> import qualified Data.Set as Set
ghci> :t Set.empty
Set.empty :: Set a
ghci> Set.empty
fromList []
ghci> Set.insert 1 (Set.insert 2 (Set.insert 3 Set.empty))
fromList [1,2,3]
```
```

File: en/lessons/basics/containers.md [hunk 64]
```diff
@@ -0,0 +1,752 @@
...
+ordering. We can see the ordering requirement from the constraints on the
+`insert` and `fromList` (and other) functions.
+
+```haskell
+ghci> :t Set.insert
+Set.insert :: Ord a => a -> Set.Set a -> Set.Set a
+ghci> :t Set.fromList
+Set.fromList :: Ord a => [a] -> Set.Set a
+ghci> :t Set.member
+Set.member :: Ord a => a -> Set.Set a -> Bool
```
Kleidukos, line 482:
```suggestion
ghci> :t Set.insert
Set.insert :: Ord a => a -> Set a -> Set a
ghci> :t Set.fromList
Set.fromList :: Ord a => [a] -> Set a
ghci> :t Set.member
Set.member :: Ord a => a -> Set a -> Bool
```

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,752 @@
...
+accessor operator.
+
+Here is an example where we want to create a container that allows us to
+retrieve ascii chars by their character code.
+
+__Note__: You can fetch the `vector` library using `cabal repl --build-depends "vector"`
+or `stack exec --package vector -- ghci`
+
+```haskell
+ghci> import qualified Data.Vector as V
```
Kleidukos, line 679:
```suggestion
ghci> import Data.Vector
ghci> import qualified Data.Vector as V
```

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,752 @@
...
+ghci> lowerCase = V.slice 97 26 asciiChars
+ghci> lowerCase
+"abcdefghijklmnopqrstuvwxyz"
+ghci> upperCase = V.slice 65 26 asciiChars
+ghci> upperCase
+"ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+ghci> nums = V.slice 48 10 asciiChars
+ghci> nums
+"0123456789"
+ ghci> V.slice 97 92 asciiChars -- Error case
```
Kleidukos, line 713: I feel like you could say a bit more than "Error case". Like "Here, we attempt <forbidden operation>, and it fails"

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,752 @@
...
+not explicit about our types.
+
+```haskell
+ghci> :set -XOverloadedLists -- This is the language extension
+ghci> [1,2,3] :: Set.Set Int
+fromList [1,2,3]
+ghci> [1,2,3] :: V.Vector Int
+[1,2,3]
+ghci> [('a', 1),('b', 2),('c', 3)] :: Map.Map Char Int
+fromList [('a',1),('b',2),('c',3)]
```
Kleidukos, line 734:
```suggestion
ghci> [1,2,3] :: Set Int
fromList [1,2,3]
ghci> [1,2,3] :: Vector Int
[1,2,3]
ghci> [('a', 1),('b', 2),('c', 3)] :: Map Char Int
fromList [('a',1),('b',2),('c',3)]
```

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,752 @@
...
+    • Illegal equational constraint GHC.Exts.Item l ~ [Char]
+      (Use GADTs or TypeFamilies to permit this)
+    • When checking the inferred type
+        it :: forall {l}.
+              (GHC.Exts.IsList l, GHC.Exts.Item l ~ [Char]) =>
+              l
+```
+
+This error is a bit of a doozey, which is why this language extension is
+typically not enabled by default. It is still good to be aware of it!
```
Kleidukos, line 752: Good thing that you show how it can fail.

Kleidukos (changes requested):
The main points of my review here are:

* Let's write down unqualified types when there is no conflict
* Let's put the import statements explicitly

I've left suggestions so that you can integrate them easily, so don't fret over the number of comments. :P

en/lessons/basics/containers.md
```diff
@@ -0,0 +1,759 @@
...
+
+It also means that we are bound to run into namespace collisions. It is common
+practice to import container modules using qualified imports. If you intend on
+following along with your own `ghci` session, you should enter the qualified
+imports below to make sure that all the modules are available in scope.
+
+```haskell
+ghci> import qualified Data.Set as Set
+ghci> import qualified Data.Map as Map
+ghci> import qualified Data.List.NonEmpty as NE
```
Kleidukos, line 36:
```suggestion
ghci> import Data.Set as Set
ghci> import Data.Map as Map
ghci> import Data.List.NonEmpty as NE
```

This works on the REPL and you have unqualified types :)

Kleidukos (changes requested):
I think we can even more optimise the imports and have 

```haskell
import Data.Foo as Foo
```
instead of two lines. Sorry for not noticing earlier :grimacing:

Kleidukos (approved):
LGTM. Godspeed! :rocket:

en/lessons/basics/containers.md
ulysses4ever: /cc @Kleidukos in case you missed this discussion about list definition.

Kleidukos: We better take this discussion on a more appropriate medium like the Slack

PR #2, opened 2021-05-27
First part of the porting to Haskell

Kleidukos:
* Remove translations
* Add a style guide alongside the glossary
* Update the README
* Rename the appropriate files


CONTRIBUTING.md
```diff
@@ -5,7 +5,7 @@ Please ensure your pull request adheres to the following guidelines:
 
 * New lessons or improvements to existing lessons are welcome.
 * Please check your spelling and grammar.
-* Please adhere to our [style guide](https://github.com/elixirschool/elixirschool/wiki/Lesson-Styleguide)
+* Please adhere to our [style guide](https://github.com/haskellschool/haskellschool/wiki/Lesson-Styleguide)
```
soupi, line 8: should be `haskellfoundation/HaskellSchool/` right?

Kleidukos: yep, good catch

style/glossary.md
```diff
@@ -0,0 +1,52 @@
...
+- [Glossary - Russian](#glossary-russian)
+  - [Elixir-related terms](#russian-elixir-related-terms)
+  - [CS terms](#russian-cs-terms)
+
+### <a name="glossary-russian"></a> Glossary - Russian
+
+#### <a name="russian-elixir-related-terms"></a> Elixir-related terms
+
+- alias - псевдоним
+- guard - ограничитель
```
ulysses4ever, line 23: (Somewhat) Commonly used Russian translation for guard is "страж" or "охранное выражение" rather than "ограничитель".

Kleidukos: Great, thank you :)

Kleidukos: If I am not mistaken, the RU glossary lists `guard` with:

```
guard (in a pattern match) | предохранитель [Виль15]; охранное выражение
```

Kleidukos: Alright, after talking with @Bodigrim, we decided that we will stick with страж, as it will give the text some fluency.

ulysses4ever: Страж is a good one. As for "предохранитель [Виль15]" I know the source and I'm pretty sure no one in practice uses this translation no matter that there was one book that used it. That wiki page with Russian translations holds a number of such practically unused translations unfortunately. But when you try to argue about that they point you to that stupid book [Виль15] 😔

Kleidukos: Yeah as someone who had to review some French translations of English terms, some authors desperately want to use their weird terminology. :)
@ulysses4ever If you want, we can pioneer new translations for these terms?

ulysses4ever: :-) I'd rather not! Страж is good enough. For the rest: it's mostly fine. I think that translations from [Виль15] are significantly off in a number of cases and better be omitted in those cases. Below are some examples but really I don't want to hold off this PR: the translation debate should probably happen somewhere else...

Here are examples where it's better to leave just one version (and many times it's not [Виль15], although there' are exceptions).

```
| ad hoc (polymorphism) | специализированный [Брон14] |
| concurrency | конкурентность [Браг13М] |
| identity function, morphism | тождественное преобразование |
| instance | экземпляр [Виль15] |
| monad transformer | монадный трансформер |
| parallelism | распараллеливание [Браг13М]; параллелизм [Виль15] |
| section (частичное применение оператора) | сечение |
| smart constructor | умный конструктор |
| strict/eager (evaluation) | строгое; энергичное [Зеф09] (вычисление) |
```

There are cases where there's no alternative but there are in fact more wide-spread versions, e.g.:

```
| bind | (монадическое) связывание |
| default definition | определение по умолчанию |
| derive (an instance) | вывести (экземпляр) |
```

One important case where there's no alternative is "list comprehensions and specifiers" (what are specifiers??). I don't think there's a wide-spread enough translation for "list comprehension", but I'm positive that "формитрователь" is not a proper Russian word, and it sounds very weird. I personally translate "list comprehension" as "генератор списка", literally "list generator". It could clash with any other notion of "generator" but luckily in Haskell we don't really have anything like that. (For Python it would cause more trouble.)

Bodigrim (approved):

PR #6, opened 2021-05-28
Add CI

JonathanLorimer:


bChiquet:
tested the deployment with tbg, this seems to work well.
Fixed to use the gh-pages url as a test.
Moved the branch this is deployed from to `main`.
Rebased.

PR #7, opened 2021-05-29
Create initial Haskell basics/basics page

### LINKED ISSUE #3: English - Basics/Basics
TOC:

* Getting Started            
    * Installing Haskell  
    * Trying Interactive Mode
* Basic Data Types           
    * Integers               
    * Floats                 
    * Booleans               
    * Strings                
* Basic Operations           
    * Arithmetic             
    * Boolean                
    * Comparison             
    * String Concatenation

TrueBoxGuy:
This resolves #3. 

I have written a page that does the following: 
* Provide instructions for installing Haskell
* Introduce `ghci` and import the `Text` type
* Cover basic data types
* Cover basic operators

### DIFF
--- a/en/lessons/basics/basics.md
+++ b/en/lessons/basics/basics.md
@@ -30,16 +30,17 @@ Follow this tutorial by running `ghci`:
     GHCi, version {{ site.ghc.version }}: https://www.haskell.org/ghc/  :? for help
     ghci>
 
-To use `Text` in Haskell, type in:
+To use the `Text` type in this tutorial, type in:
 
-```haskell
+```console?lang=haskell&prompt=ghci>,ghci|
 ghci> :set -XOverloadedStrings
 ghci> import qualified Data.Text as T
+ghci> import qualified Data.Text.IO as T
 ```
 
 You can try out Haskell by typing in a few simple expressions:
 
-```haskell
+```console?lang=haskell&prompt=ghci>,ghci|
 ghci> 2 + 3
 5
 ghci> 2 + 3 == 5
@@ -48,7 +49,7 @@ ghci> T.length "The quick brown fox jumps over the lazy dog"
 43
 ```
 
-Perhaps you're able to figure out what some of these expressions mean.
+Perhaps you are able to figure out what some of these expressions mean.
 
 ## Basic Data Types
 
@@ -56,15 +57,15 @@ Perhaps you're able to figure out what some of these expressions mean.
 
 Haskell has a bounded integer type called `Int`:
 
-```haskell
-ghci> 255 :: Int
-255
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> 196883 :: Int
+196883
 ```
 __Note__: `::` specifies the type being used.
 
 Haskell also supports arbitrary precision integers with `Integer`:
 
-```haskell
+```console?lang=haskell&prompt=ghci>,ghci|
 ghci> 123456789101112131415
 123456789101112131415
 ```
@@ -76,7 +77,7 @@ Haskell supports `Double`, a double-precision floating-point number.
 At least one digit must come before the decimal point and Haskell supports
 scientific `e` notation.
 
-```haskell
+```console?lang=haskell&prompt=ghci>,ghci|
 ghci> 3.14
 3.14
 ghci> .14
@@ -92,7 +93,7 @@ ghci> 2.7e10
 
 Booleans in Haskell are represented with `True` and `False`:
 
-```haskell
+```console?lang=haskell&prompt=ghci>,ghci|
 ghci> True
 True
 ghci> False
@@ -101,24 +102,26 @@ False
 
 ### Text
 
-Haskell supports unicode text. Strings are wrapped in double quotes:
+Haskell supports Unicode text. Text is wrapped in double quotes,
+and can be printed with `T.putStrLn`:
 
-```haskell
-ghci> "Hello" :: T.Text
-"Hello"
-ghci> "jalapeño" :: T.Text
-"jalape\241o"
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> T.putStrLn "Hello"
+Hello
+ghci> T.putStrLn "jalapeño"
+jalapeño
 ```
 
 Special characters can be represented with numeric escapes or escape codes:
 
-```haskell
-ghci> "'hi\"" :: T.Text
-"'hi\""
-ghci> "\n" :: T.Text
-"\n"
-ghci> "\42" :: T.Text
-"*"
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> T.putStrLn "'hi\""
+'hi"
+ghci> T.putStrLn "dis\njointed"
+dis
+jointed
+ghci> T.putStrLn "\42"
+*
 ```
 
 You'll learn more about data types in [collections](../collections/) and
@@ -128,10 +131,11 @@ You'll learn more about data types in [collections](../collections/) and
 
 ### Arithmetic
 
-Haskell supports the basic operators `+`, `-`, `*`, and `/`.
-Note that `/` never returns an integer.
+Haskell supports the basic operators: `+`, `-`, `*`, `/`, `**` and `^`.
+Note that `/` and `**` never operate on integers,
+whereas `^` only accepts positive integer powers.
 
-```haskell
+```console?lang=haskell&prompt=ghci>,ghci|
 ghci> 2 + 2
 4
 ghci> 2 - 1
@@ -140,11 +144,17 @@ ghci> 2 * 5
 10
 ghci> 10 / 5
 2.0
+ghci> 2 ** (-0.5)
+0.7071067811865476
+ghci> 1.5 ^ 2
+2.25
+ghci> 5 ^ 2
+25
 ```
 
 For integer division and integer modulus, use `div` and `mod`:
 
-```haskell
+```console?lang=haskell&prompt=ghci>,ghci|
 ghci> -8 `div` 3
 -2
 ghci> 243 `mod` 5
@@ -156,9 +166,9 @@ ghci> 2 * (5 `div` 2) + (5 `mod` 2)
 ### Boolean
 
 Haskell provides the boolean operators `||`, `&&`, and `not` that operate on
-`Bool`:
+`Bool`. These operators do not evaluate the second term if possible.
 
-```haskell
+```console?lang=haskell&prompt=ghci>,ghci|
 ghci> False || False
 False
 ghci> True || False
@@ -178,10 +188,10 @@ True
 ### Comparison
 
 Haskell comes with comparison operators for ordering: `<=`, `>=`, `<`, and `>`.
-The operator `==` compares equality whilst `/=` compares inequality.
+Conversely, the operator `==` compares equality whilst `/=` compares inequality.
 Values compared must be of the same type.
 
-```haskell
+```console?lang=haskell&prompt=ghci>,ghci|
 ghci> 1 > 2
 False
 ghci> 2 < 2
@@ -199,7 +209,7 @@ True
 
 String concatenation uses the `<>` operator:
 
-```haskell
-ghci> "abra" <> "cadabra" :: T.Text
-"abracadabra"
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> T.putStrLn ("abra" <> "cadabra")
+abracadabra
 ```
--- a/en/lessons/basics/basics.md
+++ b/en/lessons/basics/basics.md
@@ -1,5 +1,5 @@
 ---
-version: 1.3.0
+version: 1.0.0
 title: Basics
 ---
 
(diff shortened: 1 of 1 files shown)


en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
-After Elixir is installed, you can easily find the installed version.
+You can check the version of the Haskell compiler installed by running
+`ghc --version`:
 
-    % elixir -v
-    Erlang/OTP {{ site.erlang.OTP }} [erts-{{ site.erlang.erts }}] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]
-
-    Elixir {{ site.elixir.version }}
+    % ghc --version
+    The Glorious Glasgow Haskell Compilation System, version {{ site.ghc.version }}
```
TrueBoxGuy, line 21: Currently site.ghc.version is `8.10.4`, but the new `ghci>` prompt is used.

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
-Elixir comes with IEx, an interactive shell, which allows us to evaluate Elixir expressions as we go.
+Your installation should come with `ghci`, an interactive shell that allows you
+to try out code in real time.
 
-To get started, let's run `iex`:
+Follow this tutorial by running `ghci`:
 
-    Erlang/OTP {{ site.erlang.OTP }} [erts-{{ site.erlang.erts }}] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]
+    GHCi, version {{ site.ghc.version }}: https://www.haskell.org/ghc/  :? for help
+    ghci>
```
TrueBoxGuy, line 31: This prompt is used for consistency with `ghci` and other pages, but `>` was also considered in the documentation-task-force office hours.

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
 
-Elixir comes with IEx, an interactive shell, which allows us to evaluate Elixir expressions as we go.
+Your installation should come with `ghci`, an interactive shell that allows you
+to try out code in real time.
 
-To get started, let's run `iex`:
+Follow this tutorial by running `ghci`:
 
-    Erlang/OTP {{ site.erlang.OTP }} [erts-{{ site.erlang.erts }}] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]
+    GHCi, version {{ site.ghc.version }}: https://www.haskell.org/ghc/  :? for help
```
TrueBoxGuy, line 30: This exceeds the 80 col limit.

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
 
-```elixir
-iex> 255
+Haskell has a bounded integer type called `Int`:
+
+```haskell
+ghci> 255 :: Int
 255
 ```
+Note: specify the type to use with `::`.
```
TrueBoxGuy, line 63: This reads as if the reader must commit some action when the note merely provides information about type annotations.

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
-### Booleans
-
-Elixir supports `true` and `false` as booleans; everything is truthy except for `false` and `nil`:
-
-```elixir
-iex> true
-true
-iex> false
-false
+<interactive>:2:1: error: parse error on input ‘.’
```
TrueBoxGuy, line 84: This doesn't highlight correctly on the generated page. Furthermore, the one line gap inserted (which `ghci` does because a parse error is generated) may be confusing for the reader to parse: line gaps are often used to separate sections of code.

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
+True
+ghci> False
+False
 ```
 
-### Strings
+### Text
 
-Strings in Elixir are UTF-8 encoded and are wrapped in double quotes:
+Haskell supports unicode text. Strings are wrapped in double quotes:
```
TrueBoxGuy, line 104: Mentioning both `Text` and overloaded strings without explanation may confuse the reader. Could a better word be found?

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
-```elixir
-iex> "Hello"
+```haskell
+ghci> "Hello" :: T.Text
 "Hello"
-iex> "dziękuję"
-"dziękuję"
+ghci> "jalapeño" :: T.Text
+"jalape\241o"
 ```
```
TrueBoxGuy, line 111: The constant type annotations may be confusing. A potential solution would be to use `Data.Text.IO`, but the extra import may be confusing; this solution would also enable for one to view unicode and escape codes working.

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
-iex> :hello > 999
-true
-iex> {:hello, :world} > [1, 2, 3]
-false
-```
-
-### String Interpolation
+Haskell comes with comparison operators for ordering: `<=`, `>=`, `<`, and `>`.
+The operator `==` compares equality whilst `/=` compares inequality.
+Values compared must be of the same type.
```
TrueBoxGuy, line 182: These sentences could have a bit more cohesion. The introduction of equality and inequality seems abrupt.

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
-true
-iex> String.length("The quick brown fox jumps over the lazy dog")
+ghci> 2 + 3 == 5
+True
+ghci> T.length "The quick brown fox jumps over the lazy dog"
 43
 ```
 
-Don't worry if you don't understand every expression yet, but we hope you get the idea.
+Perhaps you're able to figure out what some of these expressions mean.
```
ulysses4ever:
```suggestion
Perhaps you are able to figure out what some of these expressions mean.
```

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
-To get started, let's run `iex`:
+Follow this tutorial by running `ghci`:
 
-    Erlang/OTP {{ site.erlang.OTP }} [erts-{{ site.erlang.erts }}] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]
+    GHCi, version {{ site.ghc.version }}: https://www.haskell.org/ghc/  :? for help
+    ghci>
 
-    Interactive Elixir ({{ site.elixir.version }}) - press Ctrl+C to exit (type h() ENTER for help)
-    iex>
+To use `Text` in Haskell, type in:
```
ulysses4ever:
```suggestion
To use the `Text` type as in examples below, type in:
```

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
+True
+ghci> False
+False
 ```
 
-### Strings
+### Text
 
-Strings in Elixir are UTF-8 encoded and are wrapped in double quotes:
+Haskell supports unicode text. Strings are wrapped in double quotes:
```
ulysses4ever:
```suggestion
Haskell supports Unicode text. Strings are wrapped in double quotes:
```

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
 
-```elixir
-iex> "Hello"
+```haskell
+ghci> "Hello" :: T.Text
 "Hello"
-iex> "dziękuję"
-"dziękuję"
+ghci> "jalapeño" :: T.Text
+"jalape\241o"
```
ulysses4ever: You may want to add that if the user wants the actual Unicode characters printed back (instead of the codepoints), they can use `putStrLn ("jalapeño" :: T.Text)`.

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
 
-    Interactive Elixir ({{ site.elixir.version }}) - press Ctrl+C to exit (type h() ENTER for help)
-    iex>
+To use `Text` in Haskell, type in:
 
-Note: On Windows PowerShell, you need to type `iex.bat`.
+```haskell
+ghci> :set -XOverloadedStrings
+ghci> import qualified Data.Text as T
+```
```
ulysses4ever, line 39: I don't buy the `Text` type for an introductory material on today's Haskell: it adds a heck lot of fuss like all those type annotations -- they will look very confusing to a novice. Someday, maybe when something like [this proposal](https://github.com/ghc-proposals/ghc-proposals/pull/409) happens, I could change my mind.

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
-iex> 2 - 1
+ghci> 2 - 1
 1
-iex> 2 * 5
+ghci> 2 * 5
 10
-iex> 10 / 5
+ghci> 10 / 5
 2.0
 ```
```
ulysses4ever, line 156: I'd add `^` and `**` which are surprisingly less known (in my experience) but are a must in today's data-centric world.

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,197 @@ Getting started, basic data types, and basic operations.
...
+ghci> 2 * (5 `div` 2) + (5 `mod` 2)
+5
 ```
 
 ### Boolean
 
-Elixir provides the `||`, `&&`, and `!` boolean operators.
-These support any types:
+Haskell provides the boolean operators `||`, `&&`, and `not` that operate on
+`Bool`:
```
ulysses4ever: Maybe worth adding a note on the usual short-circuiting behavior (Haskell is unusual in so many ways, so that this may come as a surprise).

en/lessons/basics/basics.md
TrueBoxGuy: I could go even further and print within the repl: 
>The constant type annotations may be confusing. A potential solution would be to use Data.Text.IO, but the extra import may be confusing; this solution would also enable for one to view unicode and escape codes working.

https://github.com/haskellfoundation/HaskellSchool/pull/7#discussion_r641925925

Do others agree with adding the cost of `Data.Text.IO`, and, furthermore, printing within the repl?

en/lessons/basics/basics.md
TrueBoxGuy: I'm not sure I want to introduce a datatype to beginners that I wouldn't recommend using. However, I can remove a lot of the type annotations by using functions such as `T.length` and `T.putStrLn`. 

@Kleidukos: what should I do here?

ulysses4ever: I don't understand this anxiety about `String`: if you don't write a text processing application (text as in books, novels, etc.), and, say, simply need to print a greeting to a user, why wouldn't you use `String`? Everything should be judged by the cost/benefit ratio: if your program uses just several short strings, there's no need to pull in `Text`. Note that `Char` (and therefore `String`) in Haskell is [Unicode-aware](https://www.haskell.org/onlinereport/haskell2010/haskellch6.html#x13-1190006.1.2). So we're only really trading performance here. Is this tutorial the right place to optimize for performance, or maybe we'd want to prioritize usability?

There would be a place to discuss all intricacies of all string types, I hope: `text` is just a part of the story, and somewhere we'd need to introduce ByteStrings...

Kleidukos: > if you don't write a text processing application (text as in books, novels, etc.), and, say, simply need to print a greeting to a user, why wouldn't you use String?

As a general-purpose textual data container, String is very bad. String has a very poor API, performances that go against what people know in other languages except C and the old Erlang strings, and I am afraid that if their utility disappears once we go beyond printing a greeting to a user, they shouldn't be mentioned at all.

I think linked lists should be taught as its own subject, and it should be mentioned that a common idiom is to have ad-hoc textual data represented as a list of `Char`, and outlining that it may be a nice data structure to implement templates that must be changed frequently[¹](https://samuelmullen.com/articles/phoenix-templates/).

> if your program uses just several short strings, there's no need to pull in Text.

I frequently uses several short strings and `Text` is an absolute must-have because I have _no wish_ to replicate its API on `String`. This is a matter of ergonomics and developer experience, not just data locality in contiguous slots of memory.

Kleidukos: > However, I can remove a lot of the type annotations by using functions such as T.length and T.putStrLn.

@TrueBoxGuy Let us do so then

en/lessons/basics/basics.md
Kleidukos: Agreed

en/lessons/basics/basics.md
TrueBoxGuy: Would ```To use the `Text` type in this tutorial, type in:``` be fine?

ulysses4ever: Sure.

en/lessons/basics/basics.md
Kleidukos: Can you show me how it is rendered?

TrueBoxGuy: This image
rather than this
image

TrueBoxGuy:
I've implemented these changes, but may see if there's some way to fix the highlighting. It gets problematic when printing: 
image
I've seen some libraries with repl highlighting, so it may be possible to find out a way to do this. However, these changes, along with that of adding the ghc version to `_config.yml`, should probably be part of another pull request.

en/lessons/basics/basics.md
DrewFenwick: > The operator `==` compares equality whilst `/=` compares inequality.

I'd consider rewording this. While I can catch the text's drift, a literal reading makes it sound equalities are being compared, rather than two values being compared to test for equality.

Maybe rewrite to:
> The operator `==` tests for equality whilst `/=` tests for inequality.
or
> The operator `==` checks for equality whilst `/=` checks for inequality.
---
> Values compared must be of the same type.
The phrasing seems a little bit awkward there. Consider:
> (The) compared values must have the same type.

TrueBoxGuy:
> Sorry if this has been discussed elsewhere, but wouldn't it make sense to simply hide `Prelude.putStrLn` and import `T.putStrLn` unqualified? Can be easily added in the initial blurb of imports.

@ulysses4ever 

Seems like it may be a viable idea, but I'll ask others. There was one discussion during the office hours about prelude/alternative preludes, and we came to the decision was to replicate real-life Haskell code: I don't have much experience of Haskell codebases, but how are `Text` imports for IO usually handled?

Kleidukos:
In all the codebases I've worked on, the decision was made to always use the qualified version, and I think it is honest to show this to the learner

en/lessons/basics/basics.md
TrueBoxGuy: Perhaps this is better left like this.

ulysses4ever:
I'm fine with whatever you think is better, but I want to make a meta-comment here. Using your industrial experience (e.g. with qualified imports of `Text`) when you write a beginner's tutorial may not always be the best approach to the task.

TrueBoxGuy:
I don't think the import being qualified makes it particularly more complicated (it's just a `T.` at the beginning): I agree with you that we shouldn't try to fully replicate industrial Haskell, but believe that the flip side of the spectrum is that the code shouldn't be too cute either. Having hidden name conflicts by importing Prelude without `putStrLn` seems somewhat contrived to me, and the few benefits of this are quickly lost when you use more functions from `Data.Text`. Ultimately, I think that showing the user that you can solve name conflicts by removing functions from Prelude isn't a good practice: even when writing their own non-industrial code, they'll quickly reach a point wherein using this tactic as a default offers little benefit. 

Similarly, a more important point of contention is that of using `Text`, which does introduce quite a lot of clutter to the tutorial and potentially puts off the reader. I believe that importing `Text` in this tutorial enables for the tutorial to be a baseline for the reader for the standard look of a tutorial (with the setup, such as imports and language extensions, for the tutorial being at the beginning), and somewhat removes the hurdle of parsing the structure of other tutorials. 

However, I recognise this as an unusual decision, and would be happy if a completely separate reviewer voices their opinion.

Kleidukos:
> I'm fine with whatever you think is better, but I want to make a meta-comment here. Using your industrial experience (e.g. with qualified imports of Text) when you write a beginner's tutorial may not always be the best approach to the task.

@ulysses4ever Having a qualified import of `Text` functions is far from being an "industrial" practice. Open-source projects, research projects that need proper Text functions and an appropriate data-structure will do this without hesitation. The problem here is that the Haskell standard has badly evolved, and at the present time we do what we can to compensate. Yes it sucks.

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,207 @@ Getting started, basic data types, and basic operations.
...
-```elixir
-iex> :hello > 999
-true
-iex> {:hello, :world} > [1, 2, 3]
-false
-```
-
-### String Interpolation
+Haskell comes with comparison operators for ordering: `<=`, `>=`, `<`, and `>`.
+Conversely, the operator `==` compares equality whilst `/=` compares inequality.
```
DrewFenwick:
```suggestion
Naturally the operator `==` compares equality whilst `/=` compares inequality.
```
"Conversely" means "in the converse manner", and I don't think the statement that follows it is the converse of the statement that precedes it.

You could replace it with "As you might expect" however while most users of other languages will expect `==` to check for equality, most languages don't use `/=` for inequality and it is ambiguous whether the "As you might expect" applies to both clauses.

"Accordingly" seems _more_ fitting than "conversely", but not entirely fitting either.
"Naturally" works, though it does have an effect on the tone, so feel free to use your own judgement.

DrewFenwick: > the operator `==` compares equality

Reading this literally makes it seem like the `==` operator compares equalities, rather than comparing values to see if they are equal.

You could replace "compares" with either "checks for" or "tests for".
```suggestion
Conversely, the operator `==`  checks for equality whilst `/=` checks for inequality.
```

en/lessons/basics/basics.md
```diff
@@ -9,269 +9,207 @@ Getting started, basic data types, and basic operations.
...
+Values compared must be of the same type.
```
DrewFenwick: > Values compared must be of the same type.

This reads a little awkwardly in my opinion. Consider rewording:
```suggestion
Compared values must have the same type.
```
On the other hand, to avoid a misinterpretation this might be preferable:
```suggestion
Values compared using these operators must have the same type.
```

DrewFenwick (changes requested):
I spotted just a few small wording issues.
Don't feel obliged to use my suggested changes if you can think of something better.

en/lessons/basics/basics.md
TrueBoxGuy: I like the second one quite a bit.

en/lessons/basics/basics.md
TrueBoxGuy: Yep. My changes didn't really solve https://github.com/haskellfoundation/HaskellSchool/pull/7#discussion_r641926100

TrueBoxGuy: I do want something which has a meaning close to conversely: I'm introducing equality operators now and want to highlight that contrast (that I'm covering the other side of the comparison operators). 

I feel like the word I want is "moreover": it highlights both that it's a new concept and that it adds on to what I've already said about comparison operators. What do you think?

Kleidukos (approved):
I am satisfied with the changes that have been made to address my comments.

DrewFenwick (approved):

PR #12, opened 2021-06-03
Add development CI

DrewFenwick:
Adds a Github workflow that automatically builds the `development` branch on push or PR.

This uses the same [helaili/jekyll-action@v2 action](https://github.com/helaili/jekyll-action) that the automated deployment of the main branch currently uses,
but with the `build_only` input set to `true` to avoid actually publishing.


Kleidukos (approved):
Fantastic

PR #16, opened 2021-06-09
basics/containers: Remove mention of undefined behaviour for head/tail

ulysses4ever:
Currently, basics/containers claims that `head` / `tail` leave "room for undefined behavior". [Undefined behavior](https://en.wikipedia.org/wiki/Undefined_behavior) is a technical term with a precise meaning (“result of .. a program ... is prescribed to be unpredictable”) and `head` / `tail` don't cause it ever. I think it's a very bad idea to muddle distinction between an exception (a notion of a well-defined error condition familiar from most "high-level" languages) and UB (infamously contained in C/C++ or below).


Kleidukos (approved):
Yep, good catch

PR #24, opened 2021-06-25
Create the English basics/functions lesson

### LINKED ISSUE #22: English - Basics/Functions
Lesson plan:

1. Anonymous Functions
2. Named Functions
3. Higher-order Functions
4. Local definitions

Kleidukos:
Close #22

---
image


en/index.md
```diff
@@ -1,33 +1,15 @@
 ---
 title: Elixir School
```
ulysses4ever: Should Elixir be changed to Haskell here?

en/lessons/basics/functions.md
```diff
@@ -1,425 +1,83 @@
...
-end
-
-iex> Greeter.hello ["Sean", "Steve"]
-"Hello, Sean, Steve"
+---
 
-iex> Greeter.hello ["Sean", "Steve"], "es"
-"Hola, Sean, Steve"
-```
+And that is for the basics about functions! In the next lesson, you will see that functions not only have a body, but also a type.
```
TrueBoxGuy:
```suggestion
And that is it for the basics about functions! In the next lesson, you will see that functions not only have a body, but also a type.
```

en/lessons/basics/functions.md
```diff
@@ -1,425 +1,83 @@
...
-# call with only the name key
-...> Greeter2.hello(%{name: "Fred"})
-"Hello, Fred"
-%{name: "Fred"}
-# call without the name key
-...> Greeter2.hello(%{age: "95", favorite_color: "Taupe"})
-** (FunctionClauseError) no function clause matching in Greeter2.hello/1
-
-    The following arguments were given to Greeter2.hello/1:
+Those two syntaxes exist to better let you write down your mental process. Even though it is good form to stick with one
```
TrueBoxGuy: I believe it should be `These`, because you introduced the concept.

en/lessons/basics/functions.md
```diff
@@ -1,425 +1,83 @@
...
 
-We've listed the function names in comments above.
-The first implementation takes no arguments, so it is known as `hello/0`; the second takes one argument so it is known as `hello/1`, and so on.
-Unlike function overloads in some other languages, these are thought of as _different_ functions from each other.
-(Pattern matching, described just a moment ago, applies only when multiple definitions are provided for function definitions with the _same_ number of arguments.)
+## Local definitions: `where` & `let … in`
 
-### Functions and Pattern Matching
+You can introduce bindings, such as local definition or intermediate results, through two different syntaxes:  
+The `where` keyword, and the `let … in` construct.
```
TrueBoxGuy: Is this line supposed to be rendered on a new line? It seems that the line breaks early owing to the newline in the file.

en/lessons/basics/functions.md
```diff
@@ -1,425 +1,83 @@
...
+They become particularly useful when you need to control the application of a function over a parameter:
 
-iex> Greeter2.hello()
-"Hello, anonymous person!"
-iex> Greeter2.hello("Fred")
-"Hello, Fred"
-iex> Greeter2.hello("Fred", "Jane")
-"Hello, Fred and Jane"
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> apply2 f x = f (f x)
```
TrueBoxGuy: Perhaps this function should be called `applyTwice`. I think it would be more consistent with the other function names (like `addOne`) in this document.

en/lessons/basics/functions.md
```diff
@@ -1,425 +1,83 @@
...
 3
 ```
 
-### Function Naming and Arity
+## Higher-order functions
 
-We mentioned earlier that functions are named by the combination of given name and arity (number of arguments).
-This means you can do things like this:
+We can now see a third way of using functions, which is the composition of both anonymous and named functions:
+Higher-order functions!
```
TrueBoxGuy: In British English, even if a colon is succeeded by a full sentence, you don't capitalise the succeeding clause.

en/lessons/basics/functions.md
```diff
@@ -1,425 +1,83 @@
...
+ghci> addOne 2
 3
 ```
 
-### Function Naming and Arity
+## Higher-order functions
 
-We mentioned earlier that functions are named by the combination of given name and arity (number of arguments).
-This means you can do things like this:
+We can now see a third way of using functions, which is the composition of both anonymous and named functions:
```
TrueBoxGuy: Although higher-order functions can compose anonymous and named functions, I'm not sure it's the best way of defining them.

en/lessons/basics/functions.md
```diff
@@ -1,425 +1,83 @@
...
+    │ └─ Separator between head and body of the lambda
+    └─ Declaration of the input bindigs
 ```
 
-## Named Functions
-
-We can define functions with names so we can easily refer to them later.
-Named functions are defined within a module using the `def` keyword .
-We'll learn more about Modules in the next lessons, for now we'll focus on the named functions alone.
+And incidentally, we have also seen one of the most primordial functions : The identity function, or `id`.
```
TrueBoxGuy: `:` shouldn't have a space before it, and perhaps "you have just seen" would better highlight that you're referring to `\x -> x`. 

I'm also not sure about the usage of "we" here (and in general): I haven't needed to use it in `basics.md`, but if I were to use it, I'd use it to refer to the general Haskell userbase or the writers of the article (e.g. "We introduced").

en/lessons/basics/functions.md
```diff
@@ -1,425 +1,83 @@
...
 
 {% include toc.html %}
 
 ## Anonymous Functions
 
-Just as the name implies, an anonymous function has no name.
-As we saw in the `Enum` lesson, these are frequently passed to other functions.
-To define an anonymous function in Elixir we need the `fn` and `end` keywords.
-Within these we can define any number of parameters and function bodies separated by `->`.
+Lambdas are one of the building blocks of Haskell, and their syntax is the following:
```
TrueBoxGuy:
```suggestion
Lambdas are one of the building blocks of Haskell, and they have the following syntax:
```

I think having the adjective before the noun sounds more natural here.

en/lessons/basics/functions.md
```diff
@@ -1,425 +1,83 @@
 ---
-version: 1.2.0
+version: 1.0.0
 title: Functions
 ---
 
-In Elixir and many functional languages, functions are first class citizens.
-We will learn about the types of functions in Elixir, what makes them different, and how to use them.
+Functions are at the heart of every Haskell program. Let us now see how to use them.
```
TrueBoxGuy: I think "Let us now see how to use them" lacks coherence / doesn't introduce new information to the reader. Perhaps you could talk about the nature of the article (e.g. "You'll see some basic examples of functions below").

TrueBoxGuy (changes requested):

en/lessons/basics/functions.md
Kleidukos: yep

en/index.md
Kleidukos: Curse, yes

en/lessons/basics/functions.md
Kleidukos: Thanks!

en/lessons/basics/functions.md
Kleidukos: Yeah the extra `  ` at the end of the previous line is for a newline

en/lessons/basics/functions.md
Kleidukos: Ah, didn't know that. :) Thanks!

en/lessons/basics/functions.md
Kleidukos: Ok!

soupi:
One very important aspect of functions that's worth pointing out is that they are first class values and can be treated the same way as we treat `Int` and `Bool` values (pass to functions, store in a container, etc.).

I would like to suggest:

1. Explaining that the named function syntax is a sugar for anonymous functions
2. Relocate the 'local definitions' section to a different topic that covers expressions
3. Cover recursion (?)
4. Add a little more examples to higher-order functions? Also I don't think it's correct to say that higher-order functions is the composition of anonymous functions and named functions, I would just replace the explanation with an explanation of first-class functions

en/lessons/basics/functions.md
TrueBoxGuy: I think that "The" shouldn't be capitalised here.

en/lessons/basics/functions.md
TrueBoxGuy: I think what I wanted is said more clearly here: https://github.com/haskellfoundation/HaskellSchool/pull/24#issuecomment-873465708.

TrueBoxGuy:
@soupi I agree with the majority of your points: nevertheless, it may be problematic to apply points 2 and 3 given the structure of the lessons.

For example, `let` and `where` can be seen used outside of functions in Haskell (e.g. they could be used in a request handler), but I'm not sure how this would fit into the lesson structure. 

Similarly, I think that it would be difficult to fit recursion into this lesson given that this lesson is second in the lesson plan, and pattern matching / control flow hasn't yet been covered: I think that a basic understanding of functions is necessary for later lessons.

I also believe we should try to have short lessons, but at the same time think some more examples would be useful in showing usage of the language: 
* Perhaps both named and anonymous functions should be used as arguments to the example higher order functions, and we could display another potentially interesting combinator (e.g. `on`).  
* If the section on local definitions remains, we should display that you can use the function scope and other bindings within `let` / `where`. An example where intermediate definitions are useful could be: 
```hs
fibonacciThrice a b = c + d
  where 
    c = a + b
    d = b + c
```

soupi:
I'm kind of an outsider to the project so let me share my opinion but also feel free to ignore it:

In my opinion, when writing a tutorial that splits concepts into sections (reference-like, as opposed to a more narrative based tutorial), it is best to cram most of the relevant information into the relevant section so that when someone looks for information about (for example) functions, they will find what they are looking for in the functions section, instead of having it scattered throughout the tutorial. So while it's important to keep sections short, It is more important to include the exact amount of relevant information to the section in these kinds of tutorials.

I don't know what is the structure of this tutorial currently (if there's a way to see it please link it to me) but let me suggest something else: have another section prior to the functions section called "definitions" where let/where will be covered. This section also can cover:

1. Giving names to expressions
2. Valid names / characters and common naming schemes
3. Mentioning that order of definitions does not matter
4. Mentioning that top-level expressions are not allowed
5. Explaining immutability / referential transparency (the meaning of the equals sign in Haskell)

Then the let/where part in the functions section can be remove. And instead I would suggest adding:

1. Calling functions syntax and currying + when to use parenthesis
2. Partial application
3. Function composition
4. Backticks syntax
5. Operators and precedence (how to define, and why we can write `'H' ++ replicate 9 'i'` instead of `'H' ++ (replicate 9 'i')`)

Regarding recursion, I was a bit on the fence mentioning it, and it could definitely go into a separate section dedicated to it (recursion / iteration).

TrueBoxGuy:
Yep. We could perhaps change the lesson structure, as some ideas about expressions are fundamental. 

I also believe we should comprehensively cover what would be useful to a beginner, and meant "concise" more so than "short": we shouldn't have examples which just reiterate concepts and shouldn't delve into how the language works if not necessary. 

It could also be useful to explicitly mention application syntax, parenthesis and infix functions in the functions section, but given the reference is for beginners, I'm not sure we should cover all aspects of functions, such as partial application (although I'm unable to judge how complex such a topic is given I learned Haskell in a very contrived way). This split into levels of Haskell proficiency does lead to scattering, but I'm not sure whether it's an issue given the expected reasons why developers would read the lessons (these are just my personal opinions / experience): 
* They want to write a program and figure out the equivalent way of utilising a familiar programming concept, or want to learn the syntax of Haskell. Given that the reader will know their level of proficiency, finding the appropriate lesson should be possible. 
* They want to understand a snippet of code. This is troublesome given that the titles of articles give away no information about the syntax of the concept: for example, how should a beginner figure out that `f x = x` is a function? Similarly, how should they figure out whether the syntax will be covered in the beginner section? I think we need an entirely separate solution for these users (perhaps a reference page / people that link to Haskell School when a relevant question is asked in a community).

soupi:
> I also believe we should comprehensively cover what would be useful to a beginner, and meant "concise" more so than "short": we shouldn't have examples which just reiterate concepts and shouldn't delve into how the language works if not necessary.

Overall I agree with the point, and I think these are good points in general and it's hard to find the right answer, and a lot of it depends on the main goal of this tutorial.

> given the expected reasons why developers would read the lessons (these are just my personal opinions / experience):

I would like to mention that there are already decent resources that can help learners with these type of goals. For example [The Haskell Phrasebook](https://typeclasses.com/phrasebook), [Reading Simple Haskell](https://soupi.github.io/rfc/reading_simple_haskell/) and [Haskell for Readers](http://haskell-for-readers.nomeata.de/). I would hope that Haskell School has a bit of a different goal of providing a tutorial/reference-like guide for programmers who want to understand the core building 
blocks of Haskell in a concise manner and start playing with it quickly and build things.

If that's the case, I would suggest covering things that closely related to one another together rather than splitting them by Haskell proficiency level. Especially when these things are common Haskell knowledge (or the bread an butter even). I think it would be a better fit for this kind of learning resource, but of course I might not properly understand the goals of Haskell School.

TrueBoxGuy:
I'll bring up this discussion with @Kleidukos

soupi:
A potential ToC for the functions section:

- Anonymous functions - describe the syntax (e.g. `add = \x -> \y -> x + y`),  explain that functions are first class
- Function application, currying and partial application, where to put parenthesis
- Sugars: function definition syntax, multiple parameters syntax: (e.g. `add x y = x + y` and `\x y -> x + y`)
- Higher order functions - functions are first class so they can be passed to functions, a couple of famous examples such as `map` / `sortBy` / `filter`, etc.
- Function composition - functional programming as a pipeline of data transformations

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,210 @@ version: 1.0.0
 title: Functions
 ---
 
-
+Functions are at the heart of every Haskell program, and are first-class in the language.
```
soupi: maybe elaborate what first-class means

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,210 @@ version: 1.0.0
...
 
 ## Anonymous Functions
 
-Just as the name implies, an anonymous function has no name.
-As we saw in the `Enum` lesson, these are frequently passed to other functions.
-To define an anonymous function in Elixir we need the `fn` and `end` keywords.
-Within these we can define any number of parameters and function bodies separated by `->`.
-
-Let's look at a basic example:
+Anonymos functions, or Lambdas, are one of the building blocks of Haskell. They have the following syntax:
```
soupi: Anonymos -> Anonymous

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,210 @@ version: 1.0.0
...
-
-### The & Shorthand
-
-Using anonymous functions is such a common practice in Elixir there is shorthand for doing so:
-
-```elixir
-iex> sum = &(&1 + &2)
-iex> sum.(2, 3)
-5
+ ╭─ "head", the declaration of the input bindigs
```
soupi: bindigs -> bindings, but I would suggest to avoid calling it the "head" and just call this "the input of the function"

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,210 @@ version: 1.0.0
...
+ │ 
+ │ ╭─ Separator between head and body
+ │ │ 
+ │ │  ╭─ Body of the lambda
```
soupi: I would call this "the output of a function"

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,210 @@ version: 1.0.0
...
+\x -> x
 ```
 
-As you probably already guessed, in the shorthand version our parameters are available to us as `&1`, `&2`, `&3`, and so on.
+And incidentally, you have just seen one of the most primordial functions: The identity function, or `id`.
```
soupi: maybe instead of "just seen" use "the example above is"?
I would suggest elaborating a bit more about the structure of a lambda: the input parameter can be any name, and the output any expression, the body can use any name that is currently in scope, including the parameter. for example: 

```hs
let
  greeting = "Welcome"
in
  \name -> greeting ++ " " ++ name
```

Or something similar. Maybe also mention closures at the same time? This could be further elaborated in the next section about currying (which is basically creating closures).

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,210 @@ version: 1.0.0
...
-We've listed the function names in comments above.
-The first implementation takes no arguments, so it is known as `hello/0`; the second takes one argument so it is known as `hello/1`, and so on.
-Unlike function overloads in some other languages, these are thought of as _different_ functions from each other.
-(Pattern matching, described just a moment ago, applies only when multiple definitions are provided for function definitions with the _same_ number of arguments.)
+Tacit style allows you to omit the arguments (or "points") of a function when declaring or applying it.  
+While it can be quicker and help with equational reasoning, the loss of readability that it incurs needs to be a factor
+in your decision to use it.
 
-### Functions and Pattern Matching
+For example, the function that adds 1 to a number, `(\x -> x + 1)` can be reduced to `(+1)`.
```
soupi: Consider using a function instead of an operator here, interesting candidates are `elem`, `min`, `replicate`, etc. and maybe also mention that we can only replace the last argument, so `\n -> elem n [1, 2, 3]` can't be reduced to `elem [1, 2, 3]`.

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,210 @@ version: 1.0.0
...
-    IO.puts "Hello, " <> person_name
-  end
-end
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> (\x y -> x + y) 2 3 
+5
 ```
 
-Now let's say we have a map describing a person named Fred:
+Here, the function between parentheses has been fully applied. But remember that these
```
soupi, line 79: Maybe explain what function application means (supplying values as parameters to functions), and fully applied then means "the function was supplied with all of it's arguments). Explicitly mention that supplying a value means that we remove the first layer of lambda (the first input argument) and replace (or substitute) the name with the supplied value in the body. Maybe also mention that this is formally called substitution?

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,210 @@ version: 1.0.0
...
-
-```
+This ability to do partial application unlocks many abilities, one of which allowing you to build computations whose
+parameters are not available all at the same time during the program's lifetime.
 
-The reason for this behavior is that Elixir pattern-matches the arguments that a function is called with against the arity the function is defined with.
+## Named functions
 
-Let's think about how the data looks when it arrives to `Greeter1.hello/1`:
+However, one can only go so far with anonymous functions. This is why you can give names to your functions:
```
soupi, line 96: Maybe this is a good place to talk about what first-class means, lambdas can be used just like normal data values such as Int and Bool, they can be passed to functions and returned from functions as we've seen, can be put in a data structure such as a list, and they can be *given a name* (`increment = \n -> n + 1`). There's even a syntactic sugar for the latter:

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,210 @@ version: 1.0.0
...
+3
 ```
 
-`Greeter1.hello/1` expects an argument like this:
+## Higher-order functions
 
-```elixir
-%{name: person_name}
-```
+There exists a third kind of functions, whose first argument is itself an anonymous functions: higher-order functions.
```
soupi: Maybe instead of saying that these are a third kind, just mention that this is a fancy name for functions that take other functions as input? It's not a different feature from what we've seen (as you write below), just a useful pattern that is useful in many contexts. And it doesn't have to be the first argument.

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,210 @@ version: 1.0.0
...
+Because functions are first-class components of Haskell, they are no more magical than anonymous or named functions.
```
soupi, line 107: :+1:

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,210 @@ version: 1.0.0
...
-Let's start a new function:
-
-```elixir
-defmodule Greeter2 do
-  def hello(%{name: person_name} = person) do
-    IO.puts "Hello, " <> person_name
-    IO.inspect person
-  end
-end
+ghci> filter (\x -> x > 3) [1,2,3,4,5,6]
```
soupi: maybe use `even` or `odd` for an example with regular functions? :)

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,210 @@ version: 1.0.0
...
 ```
 
-Remember that Elixir will pattern match the argument as it comes in.
-Therefore in this case, each side will pattern match against the incoming argument and bind to whatever it matches with.
-Let's take the right side first:
-
-```elixir
-person = %{name: "Fred", age: "95", favorite_color: "Taupe"}
-```
+## Local definitions: `where` & `let … in`
```
soupi, line 127: In my opinion there should be an earlier section about naming things where this should exist, should also cover indentation (how do I know where a definition ends?)

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,210 @@ version: 1.0.0
...
 ```
 
-Remember that even though it looks like `%{name: person_name} = person` is pattern-matching the `%{name: person_name}` against the `person` variable, they're actually _each_ pattern-matching to the passed-in argument.
+These two syntaxes exist to better let you write down your mental process. Even though it is good form to stick with one
+when in a collaborative setting (at work, for example), you are absolutely free to pick the one that makes you happy to
+write Haskell. :)
 
-**Summary:** Functions pattern-match the data passed in to each of its arguments independently.
-We can use this to bind values to separate variables within the function.
+## Function Composition
```
soupi, line 162: I like this section and think it is very important, I'm not a huge fan of introducing mathematical notation though, but up to you.

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,210 @@ version: 1.0.0
...
-Instead of:
+`(g ∘ f)(x)`
 
-    def foo(:first_clause, b \\ :default) do ... end
-    def foo(:second_clause, b) do ... end
+And this notation is so convenient that Haskell has kept it.
 
-one should write:
+As such, if you have two or more functions that can fit like Lego bricks, nothing more is needed than simply getting
+the arguments right. Remember the `processInt` function from earlier? It can now be written like this:
```
soupi: Consider using the `map` and `filter` examples above for the example :)

soupi (commented):
I like this, I think it covers the right topics in the right order, I do have a few suggestions for the style of the explanations. Cheers!

en/lessons/basics/functions.md
Kleidukos: I don't have a better way to show that you have a way to go from X to Y, and then from Y to Z, and their composition gives us X to Z :(

Kleidukos: But we can go back on this later.

soupi: Yeah you can just merge it and go back to it after the tutorial is pretty much complete and figure out if you want to change it

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,234 @@ version: 1.0.0
...
 ```
 
-As you probably already guessed, in the shorthand version our parameters are available to us as `&1`, `&2`, `&3`, and so on.
-
-## Pattern Matching
+And incidentally, the example above is one of the most primordial functions: The identity function, or `id`.  
+This function is very simple, as it returns precisely the argument that you pass to it.
 
-Pattern matching isn't limited to just variables in Elixir, it can be applied to function signatures as we will see in this section.
+In Lambdas, the input parameter can be any name, the output can be any expression.
```
TrueBoxGuy:
```suggestion
In Lambdas, the input parameter can be any name, and the output can be any expression.
```
This removes a comma splice.

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,234 @@ version: 1.0.0
...
-```
-
-If our function body only spans one line, we can shorten it further with `do:`:
-
-```elixir
-defmodule Greeter do
-  def hello(name), do: "Hello, " <> name
-end
-```
+The term is currification: Your function of two arguments has been transformed behind the scenes into a function
```
TrueBoxGuy:
```suggestion
The term is currification: your function of two arguments has been transformed behind the scenes into a function
```
I believe British English has slightly different capitalisation rules.

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,234 @@ version: 1.0.0
...
+[1,3,5]
 ```
 
-It finds that there is a key that corresponds to `name` in the incoming map.
-We have a match! And as a result of this successful match, the value of the `:name` key in the map on the right (i.e. the `fred` map) is bound to the variable on the left (`person_name`).
+## Local definitions: `where` & `let … in`
 
-Now, what if we still wanted to assign Fred's name to `person_name` but we ALSO want to retain awareness of the entire person map? Let's say we want to `IO.inspect(fred)` after we greet him.
-At this point, because we only pattern-matched the `:name` key of our map, thus only binding the value of that key to a variable, the function doesn't have knowledge of the rest of Fred.
+You can introduce bindings, such as local definition or intermediate results, through two different syntaxes:
```
TrueBoxGuy:
```suggestion
You can introduce bindings, such as local definitions or intermediate results, through two different syntaxes:  
```

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,234 @@ version: 1.0.0
...
-    IO.puts "Hello, " <> person_name
-    IO.inspect person
-  end
-end
+```haskell
+processInt x = timesTwo (addOne x)
+  where
+    addOne y = y + 1
+    timesTwo z = z * 2
 ```
```
TrueBoxGuy: Could you possibly show that the variable `x` can be used within the bindings?

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,234 @@ version: 1.0.0
...
+Other functions will be unable to use `addOne` and `timesTwo` since they belong to the scope of `processInt`.
 
-```elixir
-person = %{name: "Fred", age: "95", favorite_color: "Taupe"}
-```
+### `let … in`
 
-Now, `person` has been evaluated and bound to the entire fred-map.
-We move on to the next pattern-match:
+you can use its sibling, `let`, to produce the same result:
```
TrueBoxGuy:
```suggestion
You can use its sibling, `let`, to produce the same result:
```

TrueBoxGuy (changes requested):
I made a few notes about grammar in my review.

I may be able to do a more in depth review at the weekend.

en/lessons/basics/functions.md
Kleidukos: Yep, only proper nouns and acronyms are capitalised after a colon in en_UK

TrueBoxGuy (commented):
I think one of my old issues regarding capitalisation after a colon wasn't resolved.

Also, in Britain, in the context of computer science, "program" is used most commonly. I think it may be confusing to use the word "programme".

en/lessons/basics/functions.md
Kleidukos: damnit, yes

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,236 @@ version: 1.0.0
...
+(\x y -> x + y) 2
+│ 
+╰─ (\y -> 2 + y) 5
+   │ 
+   ╰─ 2 + 5
 ```
 
-`Greeter1.hello/1` expects an argument like this:
+This ability to do partial application unlocks many abilities, one of which allowing you to build computations whose
+parameters are not available all at the same time during the program's lifetime.
```
TrueBoxGuy: I feel like this would be clearer if it were split into two sentences. 

Perhaps I'm misunderstanding what is meant, but isn't this moreso a matter of convenience? If you didn't have currying, you could pass over the parameter instead of the function, although that would be clunkier (and it would make dealing with scope more of a nuisance).

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,236 @@ version: 1.0.0
...
-
-```elixir
-iex> sum = &(&1 + &2)
-iex> sum.(2, 3)
-5
+ ╭─ input of the function
+ │ 
+ │ ╭─ Separator between input and output
+ │ │ 
+ │ │  ╭─ output of the function
```
TrueBoxGuy: Could all of the annotations have their first letter capitalised?

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,236 @@ version: 1.0.0
...
+3
 ```
 
-It finds that there is a key that corresponds to `name` in the incoming map.
-We have a match! And as a result of this successful match, the value of the `:name` key in the map on the right (i.e. the `fred` map) is bound to the variable on the left (`person_name`).
+## Higher-order functions
 
-Now, what if we still wanted to assign Fred's name to `person_name` but we ALSO want to retain awareness of the entire person map? Let's say we want to `IO.inspect(fred)` after we greet him.
-At this point, because we only pattern-matched the `:name` key of our map, thus only binding the value of that key to a variable, the function doesn't have knowledge of the rest of Fred.
+There exists a third kind of functions, whose first argument is itself an anonymous functions: higher-order functions.
```
TrueBoxGuy: I agree with soupi's review here. 

Also, I believe "function" would be better suited than "functions" in some places.

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,236 @@ version: 1.0.0
...
-defmodule Greeter do
-  def hello(names, language_code \\ "en") when is_list(names) do
-    names
-    |> Enum.join(", ")
-    |> hello(language_code)
-  end
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> let addOne y = y + 1
+ghci> let timesTwo z = z * 2
+ghci> let processInt3 x = (addOne . timesTwo) x
```
TrueBoxGuy: The usage of `let` here may confuse readers.

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,236 @@ version: 1.0.0
...
+ghci> processInt3 2
+5
 
-  def hello(name, language_code \\ "en") when is_binary(name) do
-    phrase(language_code) <> name
-  end
+-- And you can have some fun
+ghci> let processInt4 x = (addOne . timesTwo . addOne . timesTwo) x
```
TrueBoxGuy: Similar thing here with `let`.

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,236 @@ version: 1.0.0
...
-    iex:15: Greeter2.hello/1
+```haskell
+processInt x = timesTwo (addOne x)
+  where
+    addOne y = x + 1
+    timesTwo z = y * 2
 ```
 
-So we've seen that Elixir pattern-matches at multiple depths because each argument matches against the incoming data independently, leaving us with the variables to call them by inside our function.
+_Note: the binding `x` is reusable throught the rest of the bindings._
```
TrueBoxGuy:
```suggestion
_Note: the binding `x` is reusable throughout the rest of the bindings._
```
I'm also not sure whether this note follows the standard formatting (I guess we'd use the one that Elixir School does). 

I like this note 🙂

TrueBoxGuy (changes requested):
I believe this lesson covers what a beginner ought to know about functions in Haskell.

en/lessons/basics/functions.md
Kleidukos: No it's not formatted, just italicised.

en/lessons/basics/functions.md
Kleidukos: Yeah you'd have to do like the JS example, with nested `returns`

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,238 @@ version: 1.0.0
...
 
-iex> Greeter.hello("Sean", "es")
-"Hola, Sean"
-```
+And this notation is so convenient that Haskell has kept it.
 
-When we combine our guard example with default arguments, we run into an issue.
-Let's see what that might look like:
+As such, if you have two or more functions that can fit like Lego bricks, nothing more is needed than simply getting
+the arguments right. Remember the `processInt` function from earlier? It can now be written like this:
```
TrueBoxGuy: I'm not sure that the analogy of lego bricks is clear: it may be easier to refer to two functions as being composable if the output type of one function matches the input type of the other function. A more complicated example with currying could also be given, such as `(applyTwice . (+)) 2 5`. 

Similarly, I'm not sure what "getting the arguments right" means: is this referring to the output function being valid if the order of the arguments is valid?

TrueBoxGuy: If the more complicated example were done, it would be ideal to explain it stepwise.

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,238 @@ version: 1.0.0
...
-
-        # 1
-        %{age: "95", favorite_color: "Taupe"}
-
-    iex:15: Greeter2.hello/1
+```haskell
+processInt x = timesTwo (addOne x)
+  where
+    addOne y = x + 1
+    timesTwo z = y * 2
```
TrueBoxGuy:
```suggestion
processInt x = timesTwo (addX x)
  where
    addX y = x + y
    timesTwo z = y * 2
```
This prevents shadowing with previous definitions and means that the variable y gets used in the function.

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,238 @@ version: 1.0.0
...
 ```
 
-So we've seen that Elixir pattern-matches at multiple depths because each argument matches against the incoming data independently, leaving us with the variables to call them by inside our function.
+_Note: the binding `x` is reusable throughout the rest of the bindings._
```
TrueBoxGuy:
```suggestion
__Note__: the binding `x` is reusable throughout the rest of the bindings.
```
This makes the formatting consistent with that of other pages (although I need to check how this formatting initially came about and it can be changed).

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,238 @@ version: 1.0.0
...
 
-If we switch the order of `%{name: person_name}` and `person` in the list, we will get the same result because each are matching to fred on their own.
+Other functions will be unable to use `addOne` and `timesTwo` since they belong to the scope of `processInt`.
```
TrueBoxGuy:
```suggestion
Other functions will be unable to use `addX` and `timesTwo` since they belong to the scope of `processInt`.
```
This follows from previous changes.

en/lessons/basics/functions.md
```diff
@@ -3,422 +3,238 @@ version: 1.0.0
...
-```elixir
-# call with same old Fred
-...> Greeter3.hello(fred)
-"Hello, Fred"
-%{age: "95", favorite_color: "Taupe", name: "Fred"}
+```haskell
+processInt2 x =
+  let addOne y = y + 1
+      timesTwo z = z * 2  
+   in timesTwo (addOne x)
```
TrueBoxGuy:
```suggestion
processInt2 x =
  let addX y = y + x
      timesTwo z = z * 2  
   in timesTwo (addX x)
```
This follows from previous changes.

TrueBoxGuy (changes requested):
Sorry, I missed a few things. 

Should be fine to merge after this though.

en/lessons/basics/functions.md
Kleidukos: So, we're going with bold instead of italics, right?

en/lessons/basics/functions.md
Kleidukos: Would this be more satisfying? 

> As such, when you have a function whose output type is the input type of another function, they are composable.

TrueBoxGuy (approved):
LGTM!

TrueBoxGuy:
I approved these changes as my example was contrived and the note about the `x` binding being reusable was sufficient.

PR #29, opened 2021-07-23
Create initial Haskell basics/operators page

### LINKED ISSUE #25: English - Basics/Operators
Make a section after #22 covering operators that does the following: 
* Strongly highlight that they are functions 
* Display how to define them 
* Show usage of backticks and parenthesis (i.e. making functions into operators and vice-versa)  
* Explain precedence

Perhaps link to a blog article on common operators (e.g. operator glossary).

TrueBoxGuy:
Closes #25. 

* Emphasise that operators are infix binary functions
* Explain how to use ordinary function identifiers as operators
* Explain usage of operators in prefix, including partial application
* Define and show examples of finding and using operator fixity
* Show a variety of custom operator and fixity declarations

### DIFF
--- a/en/lessons/basics/operators.md
+++ b/en/lessons/basics/operators.md
@@ -12,9 +12,10 @@ operators, and how infix and prefix syntax can be interchanged.
 
 ## Using Operators
 
-Operators in Haskell are functions that can be placed between two arguments: they can
-return a value of any type and their names are either a string of symbols or ordinary
-function names surrounded in backticks. Using infix notation makes it easier to write
+Operators in Haskell are functions that can be placed between two arguments.
+There are two cases where you can use a function as an operator: when the function name
+only consists of symbols or when you surround the function name in backticks.
+Using infix notation makes it easier to write
 expressions containing many common functions:
 
 ```console?lang=haskell&prompt=ghci>,ghci|
@@ -25,7 +26,6 @@ ghci> (`mod` 930233356) . negate $ 4
 ghci> - (7 + 57843)
 -57850
 ```
-__Note__: negate (`-`) is the only unary operator in Haskell.
 
 You may have seen operators such as `+`, `*`, and `.` before. The `$` operator is
 useful for removing parentheses in function application: `abs . subtract 1 $ 2 + 2` is
@@ -44,6 +44,8 @@ ghci> (+) 2 3
 ghci> (5 <=) . (* 2) $ 3
 True
 ```
+__Note__: negate (`-`) is the only unary operator in Haskell. This can make sectioning
+the subtraction operator, which uses the same symbol, difficult.
 
 The parenthesised operators in the above example can be treated as equivalent to the following
 anonymous functions:
@@ -57,6 +59,8 @@ True
 Conversely, to make a regular function identifier an operator, surround it in backticks:
 
 ```console?lang=haskell&prompt=ghci>,ghci|
+ghci> mod 2010 1998
+12
 ghci> 2010 `mod` 1998
 12
 ghci> (`div` 3) . (* 1471) $ 5
--- a/en/lessons/basics/operators.md
+++ b/en/lessons/basics/operators.md
@@ -18,30 +18,40 @@ function names surrounded in backticks. Using infix notation makes it easier to 
 expressions containing many common functions:
 
 ```console?lang=haskell&prompt=ghci>,ghci|
-ghci> abs . subtract 930233356 $ 2 + 2
+ghci> abs . subtract 119 $ 2 + 2
+115
+ghci> (`mod` 930233356) . negate $ 4
 930233352
-ghci> (`mod` 1119) . negate $ 4
-1115
 ghci> - (7 + 57843)
 -57850
 ```
 __Note__: negate (`-`) is the only unary operator in Haskell.
 
 You may have seen operators such as `+`, `*`, and `.` before. The `$` operator is
-useful for removing brackets in function application: `abs . subtract 1 $ 2 + 2` is
-equivalent to `(abs . subtract 1) (2 + 2)`.
+useful for removing parentheses in function application: `abs . subtract 1 $ 2 + 2` is
+equivalent to `(abs . subtract 1) (2 + 2)`. You can find more examples of operators on
+[FPComplete's Operator Glossary](https://www.fpcomplete.com/haskell/tutorial/operators/).
 
-### Prefix and Infix Notation
+### Interchanging Operators and Functions
 
-To use an operator symbol in prefix notation, as if it were a regular function,
-surround it in brackets. Furthermore, you can create partially applied operators by
-putting an expression on only one side of the operator:
+To use an operator symbol as if it were a regular function, surround it in parentheses.
+Furthermore, you can create partially applied operators, called sections, by putting
+an expression on only one side of the operator:
 
 ```console?lang=haskell&prompt=ghci>,ghci|
 ghci> (+) 2 3
 5
-ghci> (5 +) . (* 2) $ 8
-21
+ghci> (5 <=) . (* 2) $ 3
+True
+```
+
+The parenthesised operators in the above example can be treated as equivalent to the following
+anonymous functions:
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> (\a b -> a + b) 2 3
+5
+ghci> (\a -> 5 <= a) . (\a -> a * 2) $ 8
+True
 ```
 
 Conversely, to make a regular function identifier an operator, surround it in backticks:
@@ -49,8 +59,8 @@ Conversely, to make a regular function identifier an operator, surround it in ba
 ```console?lang=haskell&prompt=ghci>,ghci|
 ghci> 2010 `mod` 1998
 12
-ghci> (`div` 3) . (* 1471684) $ 5
-2452806
+ghci> (`div` 3) . (* 1471) $ 5
+2451
 ```
 
 
@@ -74,8 +84,8 @@ infix 4 `elem`
 
 In Haskell, operators have a precedence level between 0 and 9: this precedence level
 is the number in the operator's fixity. To see how a group of expressions
-separated by operators would be bracketed, surround the highest precedence operator
-and the expressions next to it in brackets, and repeat this process until there
+separated by operators would be parenthesised, surround the highest precedence operator
+and the expressions next to it in parentheses, and repeat this process until there
 is only one operator left at the highest level of the expression. Here's an example of
 applying this process:
 ```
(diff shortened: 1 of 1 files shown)


en/lessons/basics/operators.md
```diff
@@ -0,0 +1,164 @@
...
+### Operator Precedence and Associativity
+
+An operator's precedence and associativity combined is called its fixity, and it
+controls how the operator is parsed in an expression. To find out the fixity of an
+operator, look through `ghci`'s `:info` or the documentation for the operator:
+if fixity information isn't displayed, it defaults to `infixl 9`. The same applies to
+regular functions used in infix mode.
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> :info (+)
+infixl 6 +
```
TrueBoxGuy, line 80: Perhaps put something like `[...]` here to let the reader know text was skipped.

en/lessons/basics/operators.md
```diff
@@ -0,0 +1,164 @@
...
+16425973 + (17149685 * (106 + 235)) == 3530784511
+(16425973 + (17149685 * (106 + 235))) == 3530784511
+```
+
+Knowing operator precedence is helpful in writing code without unnecessary brackets:
+
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> (^ (3 - 1)) . subtract 3818 $ 4
+14546596
+ghci> (1729 + 1) * 121 `div` 2
```
TrueBoxGuy: Ensure that numbers are used in a way that makes the order of operations easy to deduce. For example here, the fact that the output is odd indicates that divides is of lower precedence than multiplication.

en/lessons/basics/operators.md
```diff
@@ -0,0 +1,164 @@
...
+### Prefix and Infix Notation
+
+To use an operator symbol in prefix notation, as if it were a regular function,
+surround it in brackets. Furthermore, you can create partially applied operators by
+putting an expression on only one side of the operator:
+
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> (+) 2 3
+5
+ghci> (5 +) . (* 2) $ 8
```
TrueBoxGuy: Pick operators for which the side on which the expression is would matter.

en/lessons/basics/operators.md
```diff
@@ -0,0 +1,164 @@
...
+21
+```
+
+Conversely, to make a regular function identifier an operator, surround it in backticks:
+
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> 2010 `mod` 1998
+12
+ghci> (`div` 3) . (* 1471684) $ 5
```
TrueBoxGuy: The usage of the concept of interchange may be confusing for the reader, given that this process is applied twice here: the example makes an ordinary function identifier into an operator, and then back into a prefix function that is partially applied.

TrueBoxGuy:
I'm not sure whether I've focussed enough on terminology, or whether it would get in the way of the lesson.

TrueBoxGuy:
image
Here's an image of the page.

soupi:
This is a pretty quality explanation.
I have a few comments:

- In a few places, the word "brackets" (`[]`) is used instead of "parenthesis" (`()`)
- Consider explicitly mentioning how partially applied operators are translated so the positions are more obvious. e.g.
    - `(<=)` -> `\a b -> a <= b`
    - `(1 <=)` -> `\a -> 1 <= a`
    - `(<= 1)` -> `\a -> a <= 1`

- I would suggest adding a type signature when defining an operator in at least one instance just to make sure it is covered
- Consider using this as an opportunity to discuss `++` as an example of right associativity and why it's faster than left associativity (for that case), and why `++` is slow in general (though this could be out of scope here, up to you)
- I would suggest adding a note that recommends to avoid defining operators because they make code harder to read - names usually give clues regarding what the function does and operators don't do that usually, and maybe suggest to make operators aliases to normal functions
- Consider linking to the [operators glossary](https://www.fpcomplete.com/haskell/tutorial/operators/) if it make sense

Unrelated: I'm not a fan of using ligatures and unicode symbols in code. I think this can result in confusion for the reader and make the content less accessible. An example of this is `<=` being rendered as `≤` in this document.
I suggest removing ligatures and unicode from the content.  @Kleidukos

Kleidukos:
Agreed to disable ligatures

TrueBoxGuy:
@Kleidukos do you have any plans on how we should link to FP complete? Can I just put it at the beginning of the lesson?

Kleidukos:
@TrueBoxGuy Michael allowed us to display FPCo content on Haskell School under condition of referencing the original content on the FPCo website

en/lessons/basics/operators.md
TrueBoxGuy: The text above somewhat indicates that output was skipped, but I'd like to know whether others think this is clear enough.

ulysses4ever: Adding `...` seems like a good idea to me, although not critical.

en/lessons/basics/operators.md
```diff
@@ -0,0 +1,185 @@
...
+operators, and how infix and prefix syntax can be interchanged.
+
+{% include toc.html %}
+
+
+## Using Operators
+
+Operators in Haskell are functions that can be placed between two arguments: they can
+return a value of any type and their names are either a string of symbols or ordinary
+function names surrounded in backticks. Using infix notation makes it easier to write
```
JonathanLorimer: I don't know if "they can return a value of any type" is necessary here. I think the sentence would be more clear if it read something like this:
```suggestion
Operators in Haskell are functions that can be placed between two arguments. There are two cases where you can use a function as an operator: when the function name only consists of symbols or when you surround the function name in backticks. Using infix notation makes it easier to write expressions containing many common functions:
```

en/lessons/basics/operators.md
```diff
@@ -0,0 +1,185 @@
...
+
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> abs . subtract 119 $ 2 + 2
+115
+ghci> (`mod` 930233356) . negate $ 4
+930233352
+ghci> - (7 + 57843)
+-57850
+```
+__Note__: negate (`-`) is the only unary operator in Haskell.
```
JonathanLorimer: this might be a bit confusing since the `-` is overloaded and can be used for subtraction too.

en/lessons/basics/operators.md
```diff
@@ -0,0 +1,185 @@
...
+
+Operators in Haskell are functions that can be placed between two arguments: they can
+return a value of any type and their names are either a string of symbols or ordinary
+function names surrounded in backticks. Using infix notation makes it easier to write
+expressions containing many common functions:
+
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> abs . subtract 119 $ 2 + 2
+115
+ghci> (`mod` 930233356) . negate $ 4
```
JonathanLorimer, line 24: partial application of operators might prove a bit confusing for a first example. If anything I suggest breaking this out into a different codeblock and explain what you are trying to demonstrate here.

en/lessons/basics/operators.md
```diff
@@ -0,0 +1,185 @@
...
+To use an operator symbol as if it were a regular function, surround it in parentheses.
+Furthermore, you can create partially applied operators, called sections, by putting
+an expression on only one side of the operator:
+
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> (+) 2 3
+5
+ghci> (5 <=) . (* 2) $ 3
+True
+```
```
JonathanLorimer, line 46: Ah yeah, this was the kind of explanation I was looking for in the example above.

en/lessons/basics/operators.md
```diff
@@ -0,0 +1,185 @@
...
+
+The parenthesised operators in the above example can be treated as equivalent to the following
+anonymous functions:
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> (\a b -> a + b) 2 3
+5
+ghci> (\a -> 5 <= a) . (\a -> a * 2) $ 8
+True
+```
```
JonathanLorimer, line 57: nice explanation

en/lessons/basics/operators.md
```diff
@@ -0,0 +1,185 @@
...
+
+Conversely, to make a regular function identifier an operator, surround it in backticks:
+
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> 2010 `mod` 1998
+12
```
JonathanLorimer, line 65: Maybe show `mod` used prefix first. I find the contrast can be useful for learning.

en/lessons/basics/operators.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+
+### Operator Precedence and Associativity
+
+An operator's precedence and associativity combined is called its fixity, and it
+controls how the operator is parsed in an expression. To find out the fixity of an
+operator, look through `ghci`'s `:info` or the documentation for the operator:
+if fixity information isn't displayed, it defaults to `infixl 9`. The same applies to
+regular functions used in infix mode.
```
JonathanLorimer, line 77: Can you explain what you mean by associativity? i.e. that the prefix of `r` or `l` in the `infixl 9` declaration indicates the direction in which it associates. A really good illustration of this is the `esqueleto` experimental syntax operator `:&`  which allows you to pattern match and ignore all arguments to the left because of its associativity:

```haskell
(a :& b :& c :& d :& e) == (_ :& d :& e)
```

en/lessons/basics/operators.md
```diff
@@ -0,0 +1,185 @@
...
+neutral associativity or they differ in associativity, Haskell will throw an error:
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> 29 + 61 == 3 == True
+<interactive>:1:1: error:
+    Precedence parsing error
+        cannot mix ‘==’ [infix 4] and ‘==’ [infix 4] in the same infix expression
+ghci> 0 <> 1 + (-1)
+<interactive>:2:1: error:
+    Precedence parsing error
+        cannot mix ‘<>’ [infixr 6] and ‘+’ [infixl 6] in the same infix expression
```
JonathanLorimer, line 139: nice example

JonathanLorimer (changes requested):

en/lessons/basics/operators.md
TrueBoxGuy: I agree. The fact that any function can be an operator is made clear later.

en/lessons/basics/operators.md
TrueBoxGuy: Furthermore, my note doesn't say what use knowing that "-" is also a unary operator is: I may give an example of how it makes sectioning more difficult.

en/lessons/basics/operators.md
TrueBoxGuy: You can thank soupi for it :).

en/lessons/basics/operators.md
TrueBoxGuy: Thanks. It's a small easter egg because it can typecheck if you bracket it: `(0 <> 1) + (-1) :: Sum Integer`.

en/lessons/basics/operators.md
TrueBoxGuy: Is it fine in this order?

en/lessons/basics/operators.md
TrueBoxGuy: I cover a process that explains how associativity works and then provide a couple examples later on. Do you think these examples build enough intuition? 

I feel like the most intuitive examples of associativity are in data constructors, such as that for list, but I'm not sure I should mention that here.

en/lessons/basics/operators.md
JonathanLorimer: It might be nice to briefly mention this in the intro to give context, but I leave that decision up to your discretion.

en/lessons/basics/operators.md
JonathanLorimer: Yeah, this subject does get more advanced very quickly if you delve into to much detail. Again, I leave it to your discretion.

JonathanLorimer (approved):
Looks good to me

Kleidukos (approved):
LGTM

PR #28, opened 2021-07-22
English - Basic Modules

JonathanLorimer:
Closes https://github.com/haskellfoundation/HaskellSchool/issues/26

PR #31, opened 2021-09-05
handle when there is an uncreated chapter

joelmccracken:
This fixes the issue that is currently plaguing #28 (and, I suspect, #29 as well)

---

If the file _data/contents.yml had reference to a chapter which does not
exist for the default languge, this code would error. Commenting out the
three which did not exist fixed the problem.

at time of writing, these were:
  - types
  - algebraic-data-types
  - lists-tuples
  - pattern-matching-control-flow

instead fo commenting those out though, I decided to make the code a
little more robust to handle this case.

### DIFF
--- a/_plugins/haskell_school/generator.rb
+++ b/_plugins/haskell_school/generator.rb
@@ -80,7 +80,14 @@ module HaskellSchool
             if default_lang == lang
               version_severity = 'none'
             else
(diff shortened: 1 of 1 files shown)


Kleidukos (approved):
Fantastic, thanks

PR #32, opened 2021-09-23
Add control structures and pattern matching

lbseale:
Control Structures lesson covers

- if statements
- case statements
- guards

Pattern Matching 

- Covers matching plain data constructors, lists, Maybe, Either
- Assumes that all these types have been introduced already

### DIFF
--- a/en/lessons/basics/control-structures.md
+++ b/en/lessons/basics/control-structures.md
@@ -1,207 +1,237 @@
 ---
-version: 1.1.1
+version: 1.0.0
 title: Control Structures
 ---
 
-In this lesson we will look at the control structures available to us in Elixir.
+Recall that Haskell functions are *definitions*, not lists of instructions
+
+In imperative programming, control structures are used to decide which block of
+code should be executed
+
+In Haskell, they are used instead to decide how a certain value should be
+defined
+
+__Note__: Usually control structures are used directly in the definition of a 
+function, but they can also be used in the definition of any intermediate
+value inside a function (we will see examples of both)
+
+Let's have a look at the control structures available in Haskell, and some
+examples of how they work
 
 {% include toc.html %}
 
-## `if` and `unless`
+## Extensions and Imports
 
-Chances are you've encountered `if/2` before, and if you've used Ruby you're familiar with `unless/2`.
-In Elixir they work much the same way but they are defined as macros, not language constructs. You can find their implementation in the [Kernel module](https://hexdocs.pm/elixir/Kernel.html).
+You will need to add the following lines to the top of your Haskell source 
+file (`.hs`) to be able to run these examples
 
-It should be noted that in Elixir, the only falsey values are `nil` and the boolean `false`.
+```haskell
+{-# LANGUAGE OverloadedStrings #-}
 
-```elixir
-iex> if String.valid?("Hello") do
-...>   "Valid string!"
-...> else
-...>   "Invalid string."
-...> end
-"Valid string!"
-
-iex> if "a string value" do
-...>   "Truthy"
-...> end
-"Truthy"
+import Data.Text ( Text )
 ```
 
-Using `unless/2` is like `if/2` only it works on the negative:
+## If Statements
 
-```elixir
-iex> unless is_integer("hello") do
-...>   "Not an Int"
-...> end
-"Not an Int"
+Haskell provides a simple set of keywords for `if` statements, with only a 
+single `else` condition
+
+### Syntax
+
+```
+result = if (boolExpression) 
+         then (expression1)
+         else (expression2)
 ```
 
-## `case`
+- If the first expression evaluates to `True`, then the expression after the
+  `then` keyword is evaluated
+- if it evaluates to `False`, then the expression after the `else` keyword is
+  evaluated
 
-If it's necessary to match against multiple patterns we can use `case/2`:
+### Examples
 
-```elixir
-iex> case {:ok, "Hello World"} do
-...>   {:ok, result} -> result
-...>   {:error} -> "Uh oh!"
-...>   _ -> "Catch all"
-...> end
-"Hello World"
+```haskell
+-- Is this integer even or odd?
+evenOrOdd :: Int -> Text
+evenOrOdd val = 
+    if val `mod` 2 == 0 then "Value is even" else "Value is odd"
+
+-- How to navigate a traffic light
+trafficLight :: Text -> Text
+trafficLight lightColor = 
+    if (lightColor == "Red" || lightColor == "Yellow")
+    then "Slow down"
+    else "Roll through"
+
+-- Give a nice description for the weather
+-- Note how `warmOrCold` and `niceOrGloomy` are defined in a let block then 
+-- used in the final definition
+weatherDescription :: Text -> Int -> Text
+weatherDescription weather tempInC = let
+    warmOrCold   = if tempInC > 21       then "It's warm" else "It's cold"
+    niceOrGloomy = if weather == "sunny" then "nice out"  else "gloomy out"
+    in
+    warmOrCold <> " and " <> niceOrGloomy
 ```
 
-The `_` variable is an important inclusion in `case/2` statements. Without it, failure to find a match will raise an error:
+### Usage
 
-```elixir
-iex> case :even do
-...>   :odd -> "Odd"
-...> end
-** (CaseClauseError) no case clause matching: :even
+`if` statements are not the most common control structure 
 
-iex> case :even do
-...>   :odd -> "Odd"
-...>   _ -> "Not Odd"
-...> end
-"Not Odd"
+They are handy if you have a simple definition, and don't need one of the more
+flexible structures we will see soon
+
+## Case Statements
+
+Case statements offer a way to switch on any one of a number of possible values
+
+### Syntax
+
+```
+result = case (expression) of
+    (pattern1) -> (expression1) 
+    (pattern2) -> (expression2)
+    ...
+    _          -> (expressionLast)
 ```
 
-Consider `_` as the `else` that will match "everything else".
+- The expression between the `case` and `of` keywords is evaluated
+- Its result is looked up in the list of patterns
+- For the first pattern that matches, its expression is evaluated
+- The underscore (`_`) is a catch-all pattern that will always match
 
-Since `case/2` relies on pattern matching, all of the same rules and restrictions apply.
-If you intend to match against existing variables you must use the pin `^/1` operator:
+### Examples
 
-```elixir
-iex> pie = 3.14
- 3.14
-iex> case "cherry pie" do
-...>   ^pie -> "Not so tasty"
-...>   pie -> "I bet #{pie} is tasty"
-...> end
-"I bet cherry pie is tasty"
+```haskell
+-- Offer some helpful advice, given the weather
+weatherAdvice :: Text -> Text
+weatherAdvice weatherDescription = 
+    case weatherDescription of
+        "fair"     -> "It'll be nice today"
+        "rainy"    -> "Bring an umbrella"
+        "sunny"    -> "Wear sunscreen"
+        "freezing" -> "You'll need a coat"
+        _          -> "Not sure"
 ```
 
-Another neat feature of `case/2` is its support for guard clauses:
+### Usage
 
-_This example comes directly from the official Elixir [Getting Started](https://elixir-lang.org/getting-started/case-cond-and-if.html#case) guide._
+Case statements are used commonly for intermediate values, and to decide the path
+that code will take
 
-```elixir
-iex> case {1, 2, 3} do
-...>   {1, x, 3} when x > 0 ->
-...>     "Will match"
-...>   _ ->
-...>     "Won't match"
-...> end
-"Will match"
+The *Pattern Matching* feature they use will be discussed more in the next
+lesson
+
+## Guards
+
+Guards allow you to choose between one of many possible conditions; they are
+similar to an `if`-`else if` structure
+
+### Syntax
+
+```
+result
+    | (boolExpression1) = (resultExpression1)
+    | (boolExpression2) = (resultExpression2)
+    | ...
+    | otherwise = (resultExpressionLast)
 ```
 
-Check the official docs for [Expressions allowed in guard clauses](https://hexdocs.pm/elixir/guards.html#list-of-allowed-expressions).
+- Each expression on the right side of the `|` character evaluates to a `Bool`
+- For the first one which evaluates to `True`, its expression is evaluated to
+  define the result
+- Note how the `=` symbol is not used between the result and the guards
 
-## `cond`
+### Examples
 
-When we need to match conditions rather than values we can turn to `cond/1`; this is akin to `else if` or `elsif` from other languages:
+```haskell
+-- State which value is bigger
+biggerOrSmaller :: Int -> Int -> Text
+biggerOrSmaller val1 val2
+    | val1 > val2 = "Value 1 is bigger"
+    | val2 > val1 = "Value 2 is bigger"
+    | otherwise   = "Both values are equal"
 
-_This example comes directly from the official Elixir [Getting Started](https://elixir-lang.org/getting-started/case-cond-and-if.html#cond) guide._
+-- Subjective descriptions of temperature, after converting to Celcius
+-- Note how `tmpInC` is defined in the `where` clause
+subjectiveTemp :: Double -> Text
+subjectiveTemp tmpInF
+    | tmpInC <= 0.0                   = "Freezing"
+    | tmpInC > 0.0  && tmpInC <= 20.0 = "Cold"
+    | tmpInC > 20.0 && tmpInC <= 30.0 = "Comfortable"
+    | tmpInC > 30.0                   = "Hot"
+    | otherwise                       = "Not Sure"
+  where
+    tmpInC = (tmpInF - 32.0) * ( 5.0 / 9.0 )
 
-```elixir
-iex> cond do
-...>   2 + 2 == 5 ->
-...>     "This will not be true"
-...>   2 * 2 == 3 ->
-...>     "Nor this"
-...>   1 + 1 == 2 ->
-...>     "But this will"
-...> end
-"But this will"
+-- Clothing Sizer
+-- Note that `yourSize` is defined in a where clause, then used in the
+-- definition
+whatSize :: Int -> Text
+whatSize heightInCm = "You're a size " <> yourSize
+  where
+    yourSize
+        | heightInCm > 0   && heightInCm <= 166 = "Small"
+        | heightInCm > 166 && heightInCm <= 178 = "Medium"
+        | heightInCm > 178                      = "Large"
+        | otherwise                             = "Not sure"
 ```
 
-Like `case/2`, `cond/1` will raise an error if there is no match.
-To handle this, we can define a condition set to `true`:
+### Usage
 
-```elixir
-iex> cond do
-...>   7 + 1 == 0 -> "Incorrect"
-...>   true -> "Catch all"
-...> end
-"Catch all"
+Guards are commonly used directly in function definitions, especially with
+`where` statements
+
+### Otherwise
+
+`otherwise` is an alias for `True`, and is should be used as a catch-all for the
+last guard, to ensure that one of them matches
+
+If a guard statement is evaluated, and none of its expressions return `True`, a
+runtime error occurs!
+
+#### What Happens When You Don't Use `otherwise`
+
+```haskell
+badGuard :: Text -> Text
+badGuard txtIn
+    | txtIn == "Ok" = "It worked"
 ```
 
-## `with`
-
-The special form `with/1` is useful when you might use a nested `case/2` statement or situations that cannot cleanly be piped together. The `with/1` expression is composed of the keywords, the generators, and finally an expression.
-
-We'll discuss generators more in the [list comprehensions lesson](../comprehensions/), but for now we only need to know they use [pattern matching](../pattern-matching/) to compare the right side of the `<-` to the left.
-
-We'll start with a simple example of `with/1` and then look at something more:
-
-```elixir
-iex> user = %{first: "Sean", last: "Callan"}
-%{first: "Sean", last: "Callan"}
-iex> with {:ok, first} <- Map.fetch(user, :first),
-...>      {:ok, last} <- Map.fetch(user, :last),
-...>      do: last <> ", " <> first
-"Callan, Sean"
+```console?lang=haskell&prompt=ghci>,ghci|
+*Main> badGuard "noMatch"
+"*** Exception: SchoolExamples.hs:(74,1)-(75,33): Non-exhaustive patterns in function badGuard
 ```
 
-In the event that an expression fails to match, the non-matching value will be returned:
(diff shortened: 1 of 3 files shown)


en/lessons/basics/control-structures.md
```diff
@@ -1,207 +1,237 @@
...
+-- Note that `yourSize` is defined in a where clause, then used in the
+-- definition
+whatSize :: Int -> Text
+whatSize heightInCm = "You're a size " <> yourSize
+  where
+    yourSize
+        | heightInCm > 0   && heightInCm <= 166 = "Small"
+        | heightInCm > 166 && heightInCm <= 178 = "Medium"
+        | heightInCm > 178                      = "Large"
+        | otherwise                             = "Not sure"
```
TrueBoxGuy: I like that this example shows another usage pattern for guards. 

This may be a moot point, but I don't feel that the usage pattern here matches this example, and this example is somewhat similar to the last one. I think that a more categorical example with numerical output (e.g. one which calculates scores in a game like the Prisoner's Dilemma and displays them) would be more suitable and interesting.

en/lessons/basics/pattern-matching.md
```diff
@@ -1,107 +1,268 @@
...
-```elixir
-iex> x = 1
-1
-iex> ^x = 2
-** (MatchError) no match of right hand side value: 2
-iex> {x, ^x} = {2, 1}
-{2, 1}
-iex> x
-2
+#### Examples of Not Patterns
```
TrueBoxGuy:
```suggestion
#### Examples of Non-Patterns
```

en/lessons/basics/pattern-matching.md
```diff
@@ -1,107 +1,268 @@
...
+#### Examples of Patterns
 
-When we pin a variable we match on the existing value rather than rebinding to a new one.
-Let's take a look at how this works:
+```
+"banana" -- A `Text` value
+7        -- An `Int` value
+[]       -- An empty list
+(x:xs)   -- A list with one or more values, constructed with `:`
+         -- `x` is the first element of the list, `xs` is the rest of the list
```
TrueBoxGuy:
```suggestion
         -- `x` is the first element of the list, and `xs` is the rest of the list
```

en/lessons/basics/pattern-matching.md
```diff
@@ -1,107 +1,268 @@
...
+- For the first pattern that matches the value with which the function was
+  called, its expression is evaluated
+- The underscore (`_`) is a catch-all that matches any value
+
+### Patterns In General
+
+A pattern is a combination of data constructors that could match a value
 
-## Pin Operator
+Functions are not permitted in patterns -- If you need to use a function,
```
TrueBoxGuy, line 106: I think such strong emphasis on functions not being allowed in patterns may be mildly harmful, given that view patterns and pattern synonyms exist. 

It's probably inappropriate to talk about  these language extensions here, but it may be beneficial to rephrase your explanation in terms of constructors.

en/lessons/basics/pattern-matching.md
```diff
@@ -1,107 +1,268 @@
...
-{2, 1}
-iex> x
-2
+#### Examples of Not Patterns
+
+```
+(7 - 2)     -- Functions are not allowed in patterns, `-` is a function
+(x > 5)     -- Functions are not allowed in patterns, `>` is a function
+            -- Consider a guard
+(=="apple") -- Seriously, no functions allowed, `==` is a function
```
TrueBoxGuy:
```suggestion
(7 - 2)     -- Functions are not allowed in patterns: `-` is a function
(x > 5)     -- Functions are not allowed in patterns: `>` is a function
            -- Consider a guard
(=="apple") -- Seriously, no functions allowed: `==` is a function
```

en/lessons/basics/pattern-matching.md
```diff
@@ -1,107 +1,268 @@
...
-```elixir
-iex> key = "hello"
-"hello"
-iex> %{^key => value} = %{"hello" => "world"}
-%{"hello" => "world"}
-iex> value
-"world"
-iex> %{^key => value} = %{:hello => "world"}
-** (MatchError) no match of right hand side value: %{hello: "world"}
+Using pattern matching to handle lists is very common, let's see some examples
```
TrueBoxGuy:
```suggestion
Using pattern matching to handle lists is very common. Let's see some examples
```

TrueBoxGuy (commented):
This looks great, and covers a lot of Haskell School content!

I myself don't have much to say. One pattern I may consider adding is the usage of both pattern matching and guards. Here's a contrived example of what I mean:
```hs
checkPositive (Just x) | x >= 0 = Just x
checkPositive _ = Nothing
```

en/lessons/basics/control-structures.md
lbseale: Agreed, it's not a big difference from the other example. I like your game-scoring idea, I'll think about an example for that

en/lessons/basics/pattern-matching.md
lbseale: I did not know about this language extension, so I'm learning as well. I honestly wanted a better diversity of counter-examples but I couldn't think of more than these little function applications. 

Can you think of anything other than a function application to use as a counter-example?

TrueBoxGuy: I really can't think of anything right now, but maybe something will pop into my head in the future when I'm coding. 

I think `++` is a useful counter-example (even though it's still function application).

en/lessons/basics/control-structures.md
```diff
@@ -1,207 +1,246 @@
...
-...>   {1, x, 3} when x > 0 ->
-...>     "Will match"
-...>   _ ->
-...>     "Won't match"
-...> end
-"Will match"
+- The expression between the `case` and `of` keywords is evaluated
+- Its result is looked up in the list of patterns
+- For the first pattern that matches, its expression is evaluated
+- The underscore (`_`) is a catch-all pattern that will always match
```
Kleidukos: I would love a tiny addition, like a Note, to say that if you put the `_` somewhere, every pattern that comes *after* will be ignored. Highlight that order of patterns matters!

Kleidukos (changes requested):
Thank you very much @lbseale! I'll pass the review to @TrueBoxGuy for the last part of the review and I think we can merge it afterwards ;-)

en/lessons/basics/control-structures.md
```diff
@@ -1,207 +1,246 @@
...
 
-```elixir
-iex> unless is_integer("hello") do
-...>   "Not an Int"
-...> end
-"Not an Int"
+```haskell
+{-# LANGUAGE OverloadedStrings #-}
+
+import Data.Text ( Text )
```
why-not-try-calmer, line 31: Wouldn't it be preferable to keep inside the Prelude namespace as much as possible? I know `String` are bad for many things, but don't they offer a simple and intuitive interface for teaching to new users?

en/lessons/basics/control-structures.md
```diff
@@ -1,207 +1,246 @@
...
-...> end
-"Hello World"
+```
+result = if (boolExpression) 
+         then (expression1)
+         else (expression2)
 ```
 
-The `_` variable is an important inclusion in `case/2` statements. Without it, failure to find a match will raise an error:
+- If the first expression evaluates to `True`, then the expression after the
```
why-not-try-calmer: Would it be overdoing things to recall the users with a background in dynamic languages that there is no type coercion here, so that `True` really a value constructed from the 0-place `True` constructor and not some "truthy" value as in Javascript or Python?

en/lessons/basics/control-structures.md
```diff
@@ -1,207 +1,246 @@
...
-iex> case {1, 2, 3} do
-...>   {1, x, 3} when x > 0 ->
-...>     "Will match"
-...>   _ ->
-...>     "Won't match"
-...> end
-"Will match"
+- The expression between the `case` and `of` keywords is evaluated
+- Its result is looked up in the list of patterns
+- For the first pattern that matches, its expression is evaluated
```
why-not-try-calmer: Wouldn't be nice to me more explicit: "The first pattern that matches -- and only it -- has its expression evaluated"?

en/lessons/basics/control-structures.md
```diff
@@ -1,207 +1,246 @@
...
+    | (boolExpression2) = (resultExpression2)
+    | ...
+    | otherwise = (resultExpressionLast)
 ```
 
-## `with`
-
-The special form `with/1` is useful when you might use a nested `case/2` statement or situations that cannot cleanly be piped together. The `with/1` expression is composed of the keywords, the generators, and finally an expression.
+- Each expression on the right side of the `|` character evaluates to a `Bool`
+- For the first one which evaluates to `True`, its expression is evaluated to
```
why-not-try-calmer: Wouldn't be nice to me more explicit: "The first pattern that matches -- and only it -- has its expression evaluated to"?

en/lessons/basics/pattern-matching.md
```diff
@@ -1,107 +1,283 @@
...
-## Match Operator
+## Extensions and Imports
+
+You will need to add the following lines to the top of your Haskell source 
+file (`.hs`) to be able to run these examples
+
+```haskell
+{-# LANGUAGE OverloadedStrings #-}
+
+import Data.Text ( Text )
```
why-not-try-calmer: Same remark as for the previous document above: Why make Text the default instead of String?

en/lessons/basics/pattern-matching.md
```diff
@@ -1,107 +1,283 @@
...
+
+#### Examples of Patterns
+
+```
+"banana" -- A `Text` value
+7        -- An `Int` value
+[]       -- An empty list
+(x:xs)   -- A list with one or more values, constructed with `:`
+         -- `x` is the first element of the list, and `xs` is the rest of the list
+Nothing  -- A constructor for the `Maybe` type
```
why-not-try-calmer, line 117: `Maybe` is not a type. `Maybe a` is.

en/lessons/basics/control-structures.md
Kleidukos: Reasoning about lazy linked lists when speaking of text is far from being intuitive.

en/lessons/basics/pattern-matching.md
```diff
@@ -1,107 +1,283 @@
...
+...
+functionName _ = (expressionLast)
+```
+
+- A function is defined more than one time, each with a different pattern
+- For the first pattern that matches the value with which the function was
+  called, its expression is evaluated
+- The underscore (`_`) is a catch-all that matches any value
+
+### Patterns In General
```
why-not-try-calmer, line 102: I fee llike there should be something about pattern matching on records constructor here since they are so attractive to new users, perhaps in conjunction -- unless it is deemed too advanced -- with XRecordwilcards, XRecordpuns and the 'as-' operator.

en/lessons/basics/control-structures.md
Kleidukos: Booleans are already covered in the first chapter: https://github.com/haskellfoundation/HaskellSchool/blob/4df5d05991e720cbc9255685a45388ed8dcf1900/en/lessons/basics/basics.md#booleans

en/lessons/basics/pattern-matching.md
Kleidukos: `Maybe` is very much a type, of kind `* -> *`. `Maybe a` is a type of kind `*`. Being a higher-kinded type doesn't make it less of a type.

en/lessons/basics/pattern-matching.md
Kleidukos: Interesting addition, I agree.

en/lessons/basics/control-structures.md
why-not-try-calmer: So this document is solely about syntax, and the syntax/interface provided by Strings is extremely intuitive. It's only their semantics that is not. But since the semantics of Haskell in general is not intuitive at all, why not handle semantic difficulties separately?

en/lessons/basics/pattern-matching.md
why-not-try-calmer: What I mean to say is that since `Nothing :: Maybe a`, the correct thing to say is that `Nothing` is a constructor for the `Maybe a` datatype. More generally, and to put the point in different terms, C is a constructor to a type T only if C constructs of type whose kind = the kind of T. Since `Nothing` and `Maybe` does not agree in kind, the former cannot be a constructor to the latter.

en/lessons/basics/control-structures.md
Kleidukos: Because then you have to undo the intuition that you promote afterwards. We already have enough space leak issues with people blindingly using `String` because they are "so intuitive to use", I don't want to contribute legitimising `String` as a valid textual datatype: It is not and we should stop fooling ourselves.

lbseale: I noticed that the very first lesson uses `Text`, and I figured that was intentional. So I did the same here. Also using the `<>` operator is just as intuitive to me as using `++` with `String`

en/lessons/basics/pattern-matching.md
lbseale: Nice one, I'll add this

en/lessons/basics/control-structures.md
why-not-try-calmer: > I noticed that the very first lesson uses `Text`, and I figured that was intentional. So I did the same here. Also using the `<>` operator is just as intuitive to me as using `++` with `String`

I very much understand the idea of promoting the `text` package. Everyone doing anything serious in Haskell uses it instead of `String` for almost anything regarding text contents -- no question about that. Also `Data.Text` might become part of the Prelude one day (after we move from Utf16 to Utf8 probably). But if you look at any lauded pedagogical material for Haskell (_Haskell Programming From First Principles_, more controversially _Get Programming With Haskell_ for example), the `Text` type is introduced _much later_ and for reasons. (quoting from Kurt Manning:

> While learning Haskell, String is useful for two reasons. First, as mentioned, many of the basic string utilities are baked into the standard Prelude. Second, lists are to Haskell what arrays are to C. Many concepts in Haskell are nicely demonstrated with lists, and strings are useful lists. For learning purposes, feel free to stick with String.

To me the first reason means: now you need to `import Data.Text` everywhere, which clutters the material, while the second reason means that you can easily use String to illustrate lists -- including the list monad. I think these are two valid and strong reasons for the sake of pedagogy.

And there are other reasons; now you need to explain why one should import the strict and not the lazy version, or vice-versa; why `Data.Text` has its own separate functions different from the Prelude functions; and if the point is efficient text processing, why `Data.Text` instead of `ut8-bytestring` or `Data.Vector.Unboxed.Vector Char`, and why --- exception! -- `Text.IO` is awfully inefficient and should be avoided.

In other word, it seems to me that using `Text` everywhere is a bit of a premature optimization that opens a can of worms. How about instead:

1) using String whenever the focus is on cons-able data structures?; and
2) using Text whenever the focus is on efficient text processing?

Kleidukos: Well, you're entitled to your opinion. :)

why-not-try-calmer: > Well, you're entitled to your opinion. :)

Can we get the perspective of the folks supporting the Haskell School project on this? It seems to be important, both as a technical detail and as a case illustrating the vision of this project. It doesn't feel right to discard the dozens of examples of user-friendly books and documentation about Haskell by implicitly telling them that Text is better than String for the purpose of learning Haskell.

TrueBoxGuy: Haskell School does need to keep its own style (as a reference material to get people quickly acquainted with Haskell). 

I don't feel like it's particularly productive to debate over the usage of `Text`: it's a project wide decision and debating over it slows down the progress of individual pull requests. It would be better to get to an initial release and then see whether the benefits of using `Text` have been realised. 

However, these internal decisions perhaps aren't the clearest from the outside and your review has been very helpful in showing potential misunderstandings a reader may have.

why-not-try-calmer: Sure, I don't want to slow down this PR with a project-wide decision. Shall I open an Issue then?

TrueBoxGuy: I think we're considering putting these decisions in the style guide, and then maybe there can be a discussion about the style guide somewhere.

ulysses4ever: @why-not-try-calmer I share the concern of using `text` from the get-go, [as I mentioned in an earlier PR](https://github.com/haskellfoundation/HaskellSchool/pull/7#discussion_r642747500).

I encourage Haskell School government to seek a wider input on this matter, e.g. via Discourse / Reddit / mailing lists.

lbseale: I support teaching `Text` from the beginning. I learned Haskell in the last year, and it feels bad to learn topics as a beginner, only to find out that they're actually bad practice. I think there are several examples of this

1. `String` - you find out that the tools for it are clunky and that you should use `Text` instead
2. Lists - You find out that usually you should use some `Map` type instead, and that lists involve unsafe partial functions you're not supposed to use
3. Explicit recursion - You find out it can cause memory leaks, and that you rarely actually use it in practice

There are probably others. TBH these issues are why I want to contribute to this project.

Kleidukos: I am going to speak as the Editor-in-chief of this content platform: 

Teaching `Text` from the beginning has several advantages:
* It helps reducing cognitive load, as the learner doesn't have to "un-learn" String
* It gives the learner proper habits from the beginning
* It reduces the amount of confusion arising from the fact that `String` is simply an alias for `[Char]` and GHC doesn't always expand or retain aliases.
* The `String` utilities baked into the Prelude are simply insufficient when going beyond very simple or contrived examples, and they possess equivalents in `Text`.

As a general statement, `String` is a historical accident that we kept and I do not wish to promote something from which languages older than Haskell like Erlang have moved away.

@why-not-try-calmer @ulysses4ever You may not agree with this decision but such is life. @TrueBoxGuy is editor for the English language part of Haskell school, and I (and transitively, the executive team of the Haskell Foundation) have put trust in his judgement.

JonathanLorimer: FWIW I am strongly in favour of using `Text` over `String`.

Perhaps it is our fault for not being clear enough in some coding style document, but one of the primary motivating factors of this educational tool is to get a new haskeller up to speed and working with haskell in real world contexts quickly. To this end we did not want to delay learning to use `Text` over `String`. We also don't want to delay introducing them to the critical haskell libraries that they will inevitably need to learn. Also, if the argument is that `String` is convenient for teaching pattern matching, surely any other example of a list that is not an overloaded concept would be an equally good example?

Haskell School seeks to learn from mistakes made by other educational haskell materials (like the ones you mention @why-not-try-calmer). If you don't consider them to be mistakes, then perhaps I can appeal to the virtues of having diversity in available resources.

Finally, I would like to make an appeal to authority. @Kleidukos was designated the shepherd of this project, and TBG / myself have put in hours of time, early on, contributing to the project. All three of us came to the conclusion that using `Text` throughout is the approach we think is best.

I don't think this project should be an autocracy, if @ulysses4ever (who has also been around this project and contributing since the beginning) and `why-not-try-calmer` want to make a case for teaching `String` over `Text` I encourage you to bring it up in the Haskell Foundation slack and we can have it out there. But I hope its clear that this was not a one off decision, it is in service of our larger vision for the project. To summarize I think we should stick with the current approach since:

1. We have made the decision to prioritize getting the learner doing real world things as quickly as possible. This means introducing them to libraries outside of prelude earlier than other pedagogical materials
2. We think that, at best, we are improving on the canon of haskell educational material. At worst, we are providing diversity, which I argue is an end in its own right.
3. This decision was arrived at by people who were charged with executing the project, and who have invested their time in bringing it about. Know that we have discussed this in the past, and our decision was made based on what we understand the mandate of the project to be.

JonathanLorimer: Okay, it seems @Kleidukos and I posted at almost exactly the same time. I hope this doesn't come across as needless dog piling. Just wanted to clarify that.

shadowcat-mst: Representing strings as List-of-Char, especially in a unicode world, is one of the most common initially enticing decisions for all functional programming language implementors that in the long run turn out to be where sanity and performance go to die. 

I think Erlang mostly had it because of starting off being implemented in prolog, which really mostly didn't care about string handling as originally designed.

Any "let's implement a lisp" tutorial that's written by somebody who's worked with non-toy systems will almost certainly contain a paragraph to the effect of "we know this looks like a really tempting idea but trust us, don't".

It's unfortunate that String already stole a bunch of operator names that one might've wished one could use for Text, but that's the nature of long lived systems, and teaching newbies Text first will be a slightly steeper learning curve up front in return for much *much* less pain for them later, so honestly starting them off with String in 2021 would just be unnecessary cruelty.

Yours, mst, speaking from bitter experience over multiple FP implementations.

why-not-try-calmer: Fair enough, this PR is probably not the best place for debating the presupposition -- which by now you'll know I don't accept -- that illustrating basic Haskell with `String`s entails a general endorsement or promotion of `String`s. Feel free to ignore the initial comment that spawned this discussion.

en/lessons/basics/pattern-matching.md
lbseale: Just added this to latest push

en/lessons/basics/pattern-matching.md
lbseale: Decided to leave this as-is

en/lessons/basics/control-structures.md
lbseale: Added to latest push

en/lessons/basics/control-structures.md
lbseale: Added to latest push

en/lessons/basics/control-structures.md
lbseale: I added a small comment to be more explicit about this, and to be consistent with the section on Guards

en/lessons/basics/pattern-matching.md
```diff
@@ -1,107 +1,426 @@
...
+
+Two language extensions can add convenience to working with records and patterns
+
+1. `RecordWildCards` can automaticlly define local names that exactly match the
+   record field names
+2. `NamedFieldPuns` allows you to just input the record field name, and have it
+   defined as a local name
+
+Feel free the explore these further once you feel confident writing record
+patterns by hand
```
Kleidukos: A period is missing. :)

Kleidukos (changes requested):
Nitpick. :)

en/lessons/basics/pattern-matching.md
lbseale: None of the one-line statements have periods. I think it looks a bit nicer this way, should I go and add periods to them?

Kleidukos: I think that from a typographical point of view it's better, yes. :)

lbseale: Ok just pushed with added punctuation 👍

Kleidukos (approved):
Well it would seem we've arrived to a satisfying point! Thank you very much @lbseale for your patience and dedication. :)

PR #41, opened 2021-11-05
Add Basics/Types

lbseale:
Add types module

Covering (for now)
- Static Type
- Type Inference
- Bools, Ints, Floats, Chars, Text
- Why we don't use String

This is just the start, I plan to add more details

## Submitter checklist

- [x] My PR is related to Issue #38 
- [x] My PR follows the [code style](https://github.com/haskellfoundation/HaskellSchool/blob/development/style/code.md)
- [x] My PR follows the [style guide for my language](https://github.com/haskellfoundation/HaskellSchool/tree/development/style)
- [x] I have read and understood the [CONTRIBUTING guide](https://github.com/haskellfoundation/HaskellSchool/blob/development/CONTRIBUTING.md)

### DIFF
--- a/en/lessons/basics/types.md
+++ b/en/lessons/basics/types.md
@@ -1,0 +1,177 @@
+---
+version: 1.0.0
+title: Types
+---
+
+Types are a fundamental and integral part of Haskell. Haskell's advanced 
+type system sets it apart from other languages, and is a powerful method for
+ensuring that your code does what you want it to do.
+
+{% include toc.html %}
+
+## Extensions and Imports
+
+You will need to add the following lines to the top of your Haskell source 
+file (`.hs`) to be able to run these examples:
+
+```haskell
+{-# LANGUAGE OverloadedStrings #-}
+
+import Data.Text ( Text )
+```
+
+## Static Typing
+
+Haskell is a statically-typed programming language. This means that the type
+of all data must be known before the program is compiled. Some imperative 
+languages, such as Java and C++ are also statically-typed, while others like 
+Python and JavaScript are not.
+
+## Type Inference
+
+Haskell's type system works very differently from that of Java and C++. 
+Instead of declaring the type of every single function and variable, Haskell
+can *infer* the type of values based on just a few declarations.
+
+Typically, you just need to specify the type of a function. The types of all
+intermediate values used in the function will be inferred by Haskell, and you
+don't have to state their types explicitly.
+
+Type inference is a killer feature of Haskell, and once you get used to it, you
+won't want to work without it.
+
+## Type Declarations
+
+In Haskell, you declare types on a separate line from where they are used. 
+Even though I said above that you usually only declare types for functions, in
+these examples we are going to declare them for single values.
+
+Type names are always capitalized in Haskell, and value (or function) names 
+are always lower-case.
+
+### Syntax
+
+```
+valueName :: Type
+valueName = definition
+```
+
+## Primitive types
+
+### Boolean
+
+Haskell has booleans, and they can either be `True` or `False`
+
+```haskell
+imTrue :: Bool
+imTrue = True
+
+imFalse :: Bool
+imFalse = False
+```
+
+### Integer
+
+Haskell has two common integer types, `Int` and `Integer`
+
+#### `Int`
+
+A fixed-precision type, with at least the range `[-2^29 .. 2^29-1]`, or 30 bits
+It is commonly implemented as 32 or 64 bits
+
+```haskell
+positiveInt :: Int
+positiveInt = 42
+
+negativeInt :: Int
+negativeInt = -37
+```
+
+The fact that `Int`'s size is limited means it is subject to overflows. These 
+occur without warning, so if they are a concern, use `Integer`
+
+#### `Integer`
+
+An arbitrary-precision type. Its size is only limited by the memory on your
+machine.
+
+```haskell
+positiveInteger :: Integer
+positiveInteger = 42
+
+negativeInteger :: Integer
+negativeInteger = -37
+```
+
+### Floating Point
+
+Haskell has floating-point numbers, single-precision `Float` and 
+double-precision `Double`
+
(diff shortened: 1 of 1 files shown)


en/lessons/basics/types.md
```diff
@@ -0,0 +1,177 @@
...
+machine.
+
+```haskell
+positiveInteger :: Integer
+positiveInteger = 42
+
+negativeInteger :: Integer
+negativeInteger = -37
+```
+
```
Kleidukos, line 105: Any chance you could add a little something about `Word`? I'm especially interested in promoting it as a type that can't be negative, instead of using `Int` and hoping for the best. It's also the occasion to evoke the *correct by construction* paradigm.

en/lessons/basics/types.md
```diff
@@ -0,0 +1,177 @@
...
+### Text
+
+Text values can go in the `Text` type
+
+`Text` literals are always surrounded with double quotes (`""`)
+
+```haskell
+someText :: Text
+someText = "Let's learn Haskell"
+```
```
Kleidukos, line 160: Could you add a couple of examples for `Text`? Especially the functions that do not exist in `base` for `String`?

Kleidukos (changes requested):

en/lessons/basics/types.md
lbseale: Yes, great idea

en/lessons/basics/types.md
lbseale: 👍

soupi: Consider mentioning `Natural` instead of or together with `Word`?

en/lessons/basics/types.md
```diff
@@ -0,0 +1,177 @@
...
+```haskell
+{-# LANGUAGE OverloadedStrings #-}
+
+import Data.Text ( Text )
+```
+
+## Static Typing
+
+Haskell is a statically-typed programming language. This means that the type
+of all data must be known before the program is compiled. Some imperative
```
soupi, line 26:
```suggestion
of all data must be known before the program is executed. Some 
```

en/lessons/basics/types.md
tchoutri: Yes!!

Kleidukos: And the fact that arithmetic underflows are handled differently:

When doing `0 - 1`

* Since `Word` is modular, it loops back to `18446744073709551615`
* In the case of `Natural`, an `arithmetic underflow` exception is thrown

en/lessons/basics/types.md
```diff
@@ -0,0 +1,177 @@
...
+languages, such as Java and C++ are also statically-typed, while others like 
+Python and JavaScript are not.
+
+## Type Inference
+
+Haskell's type system works very differently from that of Java and C++. 
+Instead of declaring the type of every single function and variable, Haskell
+can *infer* the type of values based on just a few declarations.
+
+Typically, you just need to specify the type of a function. The types of all
```
soupi, line 36: Since this isn't actually required, consider saying something like this instead:

 while Haskell does not require type annotations in most cases, it is good practice to annotate the types of top-level definitions. This adds an extra layer of documentation, and safety where the Haskell compiler can verify our intent (as expressed in the type annotation) matches the code.

en/lessons/basics/types.md
```diff
@@ -0,0 +1,177 @@
...
+intermediate values used in the function will be inferred by Haskell, and you
+don't have to state their types explicitly.
+
+Type inference is a killer feature of Haskell, and once you get used to it, you
+won't want to work without it.
+
+## Type Declarations
+
+In Haskell, you declare types on a separate line from where they are used. 
+Even though I said above that you usually only declare types for functions, in
```
soupi, line 46: You can avoid this line by saying "top-level definitions"

en/lessons/basics/types.md
```diff
@@ -0,0 +1,177 @@
...
+can *infer* the type of values based on just a few declarations.
+
+Typically, you just need to specify the type of a function. The types of all
+intermediate values used in the function will be inferred by Haskell, and you
+don't have to state their types explicitly.
+
+Type inference is a killer feature of Haskell, and once you get used to it, you
+won't want to work without it.
+
+## Type Declarations
```
soupi, line 43: Consider changing this to "type signatures" or "type annotations", so it won't be confused with declaring new types (using `data`/`newtype`/etc)

en/lessons/basics/types.md
```diff
@@ -0,0 +1,177 @@
...
+```haskell
+positiveInt :: Int
+positiveInt = 42
+
+negativeInt :: Int
+negativeInt = -37
+```
+
+The fact that `Int`'s size is limited means it is subject to overflows. These 
+occur without warning, so if they are a concern, use `Integer`
```
soupi, line 91:
```suggestion
occur without warning, so if they are a concern, use `Integer`.
```

en/lessons/basics/types.md
```diff
@@ -0,0 +1,177 @@
...
+
+You will need to add the following lines to the top of your Haskell source 
+file (`.hs`) to be able to run these examples:
+
+```haskell
+{-# LANGUAGE OverloadedStrings #-}
+
+import Data.Text ( Text )
+```
+
```
soupi, line 22: Maybe add "we'll explain why this is needed later" or something similar?

en/lessons/basics/types.md
```diff
@@ -0,0 +1,177 @@
...
+is a linked-list of single characters. While this may be a satisfying use of
+linked lists, it is not a practical design for representing strings.
+
+Since it has been around from the beginning, and it works fine for small
+strings, the `String` type is common in Haskell. It is used everywhere in
+Haskell's default libraries.
+
+`Text` is more convenient, performant and capable than `String`. We want 
+beginners to avoid the mistakes of the past and start off with good 
+practices, so we have chosen to teach `Text`.
```
soupi, line 177: Maybe this would be a good place to mention that text lives in a different package/module, what overloadedstrings does, and how to convert String <-> Text?

PR #42, opened 2021-11-27
Fix nixpkgs version in shell.nix

blackheaven:
Jekyll installation was breaking on my machine because the version in `shell.nix` wasn't fixed.


Kleidukos (approved):
LGTM, thanks!

PR #45, opened 2022-02-21
Guidelines for writing about a typeclass

chris-martin:
As I'm getting the first typeclass lesson wrapped up, from it I've put together a list of the things I think are important to cover in each lesson, specifically when teaching a typeclass. Perhaps it is of general utility to include here in the style guide?

### DIFF
--- a/style/code.md
+++ b/style/code.md
@@ -26,5 +26,17 @@ import qualified Data.Vector as V
 * For terminal commands, _shell_ code blocks must be used, and the prompt should be `$`
 * Keep examples simple when possible
 
+## Typeclasses
 
+Checklist for writing about a typeclass:
 
+* What types belong to the class? What types do not?
+* Give an example of using at least one of the class methods directly.
+* In addition to the class methods themselves, what are some other examples of library functions that include this class as a constraint? Demonstrating a task that is made easier by having this abstraction is often a critical part explaining the value of the typeclass. For example, `Ord`'s `compare` gives rise to `sort`, and `Semigroup`'s associative `(<>)` is what makes `stimes` possible.
+* How are instances of this class typically obtained? Written by hand, or derived? If derived, by what strategy? Include examples for strategies that are commonplace.
+
+Exercise ideas for lessons about a typeclass:
+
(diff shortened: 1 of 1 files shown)


style/code.md
```diff
@@ -26,5 +26,17 @@ import qualified Data.Vector as V
 * For terminal commands, _shell_ code blocks must be used, and the prompt should be `$`
 * Keep examples simple when possible
 
+## Typeclasses
 
+Checklist for writing about a typeclass:
 
+* What types belongs to the class? What types do not, any why?
```
TrueBoxGuy:
```suggestion
* What types belong to the class? What types do not, and if not why?
```
I think this reads more clearly.

TrueBoxGuy (commented):
This looks perfect! 

There might be no need to explicitly mention it here, but it would be nice if articles referred to the value of the typeclass when not immediately obvious (e.g. it might abstract over a certain form of computation / make some form of computation easier). 

For example, the associative nature of a semigroup allows for the method of repeating squares to be used in `stimes`, which makes it easy to efficiently compute powers of matrices, numbers (modulus some value), permutations, etc.

TrueBoxGuy:
Also, this definitely requires further discussion, but with the nature of writing of the Haskell School somewhat tending towards quick-start documentation, it might make sense to present exercises differently. 

It's probably still useful to label the exercises and give the reader time to think them through but maybe they could be framed and explained more like examples (i.e. with the article walking through the thought process).

chris-martin:
> There might be no need to explicitly mention it here, but it would be nice if articles referred to the value of the typeclass when not immediately obvious (e.g. it might abstract over a certain form of computation / make some form of computation easier).
> 
> For example, the associative nature of a semigroup allows for the method of repeating squares to be used in `stimes`, which makes it easy to efficiently compute powers of matrices, numbers (modulus some value), permutations, etc.

Exactly, that's what I was getting at. If you didn't already think it was explicitly mentioned here, I'll take that as a queue to say a bit more -- updated to elaborate by way of borrowing a few of your words.

Kleidukos:
@chris-martin hi! Any update on this PR? :)

chris-martin:
> @chris-martin hi! Any update on this PR? :)

I believe that force-push did make the changes I mentioned. This is good to go, I think?

Kleidukos (approved):

Kleidukos:
Perfet :)

PR #46, opened 2022-02-22
Eq and Ord

chris-martin:
First typeclasses lesson, `Eq` and `Ord`! Love to know what y'all think - If this looks good, we're ready to put the rest of this project on the front burner for a while.

### DIFF
--- a/en/lessons/typeclasses/comparison.md
+++ b/en/lessons/typeclasses/comparison.md
@@ -57,8 +57,8 @@ error:
 Suppose we define a `Message` as a product of two [`Text`](https://hackage.haskell.org/package/text/docs/Data-Text-Lazy.html#t:Text) values:
 
 ```haskell
-import Data.Text.Lazy (Text)
-import qualified Data.Text.Lazy as Text
+import Data.Text (Text)
+import qualified Data.Text as Text
 
 data Message = Message { subject :: Text, body :: Text }
 
@@ -235,8 +235,8 @@ This class has two methods: `(==)` and `(/=)`. The `MINIMAL` pragma informs us t
 In this following example, we define a datatype called `CaseInsensitiveText`. It is like `Text`, but with an equality test that disregards differences in capitalization of letters.
 
 ```haskell
-import Data.Text.Lazy (Text)
-import qualified Data.Text.Lazy as Text
+import Data.Lazy (Text)
+import qualified Data.Lazy as Text
 
 data CaseInsensitiveText = CaseInsensitiveText{ caseSensitiveText :: Text }
 
@@ -511,8 +511,8 @@ The `Ord` class has seven methods. You can define all of these for your type, bu
 Consider the `CaseInsensitiveText` example from before. Since we customized `Eq`, we should also write a custom `Ord` instance.
 
 ```haskell
-import Data.Text.Lazy (Text)
-import qualified Data.Text.Lazy as Text
+import Data.Text (Text)
+import qualified Data.Text as Text
 
 data CaseInsensitiveText = CaseInsensitiveText{ caseSensitiveText :: Text }
 
--- a/en/lessons/typeclasses/comparison.md
+++ b/en/lessons/typeclasses/comparison.md
@@ -54,7 +54,7 @@ error:
 
 ### Deriving Eq for new types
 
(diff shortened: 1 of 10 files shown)


en/lessons/typeclasses/comparison.md
```diff
@@ -0,0 +1,590 @@
...
+False
+ghci> message1 /= message2
+True
+```
+
+Obtaining an `Eq` instance in this way is called "stock deriving" because it takes advantage of the compiler's built-in support for deriving the `Eq` class.
+
+#### When Eq cannot be derived
+
+To derive a stock `Eq` instance for a product type like `Message`, all of the constructor's fields must have `Eq` instances. Suppose our type contains an additional field, an `IO` action which sends the message somewhere.
```
Kleidukos: I think I would phrase it like

"For a record named `Message`". 

I fear "product type" may lead the reader to think about the wrong things. Inversely, if the subject was cardinality of types, I'd definitely use "product type"!

en/lessons/typeclasses/comparison.md
```diff
@@ -0,0 +1,590 @@
...
+```
+
+(When we discuss the `Ord` class below, we will see a faster alternative to `nub` that requires comparisons of fewer element pairs.)
+
+##### Exercise: No instance ... arising from a use of ...
+
+The following code does not compile. Try to compile it, and read the ensuing error message.
+
+```haskell
+import qualified Data.List as List
```
Kleidukos, line 180: Awesome, I love this style of import.

en/lessons/typeclasses/comparison.md
```diff
@@ -0,0 +1,590 @@
...
+
+This class has two methods: `(==)` and `(/=)`. The `MINIMAL` pragma informs us that an instance must provide a definition for either `(==)` or `(/=)`.
+
+#### Case-insensitive text
+
+In this following example, we define a datatype called `CaseInsensitiveText`. It is like `Text`, but with an equality test that disregards differences in capitalization of letters.
+
+```haskell
+import Data.Text.Lazy (Text)
+import qualified Data.Text.Lazy as Text
```
Kleidukos: Any particular reason why lazy Text is used here?

Kleidukos (changes requested):
Thank you very much for the PR!
I have a couple of remarks about the "comparison" lesson, but that's really it. :)

en/lessons/typeclasses/comparison.md
chris-martin: Wouldn't mind changing it, but - I use "product" because it sounds related to "sum", and these two varieties of type are often discussed together in contrast. If we avoid saying "product" in favor of "record", is there also an alternative to "sum"?

en/lessons/typeclasses/comparison.md
chris-martin: A reason, but not one I'm extremely confident in. I know a lot of people are frustrated by having to choose between strict and lazy text before they really know how to decide which is more appropriate, so I figure Haskell School should consistently recommend a primary go-to standard choice, if one can be agreed upon. My rough perception of popular sentiment is that people prefer APIs that can work with lazy text, since lifting from strict to lazy for compatibility with some library is a cheaper operation than forcing lazy to strict. But it is only a rough sense, and perhaps it would be good to hear some other voices before doubling down on it.

en/lessons/typeclasses/comparison.md
Kleidukos: I think we can safely express the "intent" of a sum type by calling it a "variant".

Kleidukos: Other languages call them "tagged unions" as well, but I'm more in favour of "variant".

en/lessons/typeclasses/comparison.md
Kleidukos: Yes I see what you mean.  
The reason is good, but my decision would have been different. I would go with strict text by default, and use a proper streaming primitive when it is needed.

Kleidukos:
@chris-martin Hi! any update on this PR? :)

chris-martin:
Made requested changes!

Kleidukos (approved):
Thank you :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #48, opened 2022-06-13
Enum and Bounded

chris-martin:
This lesson covers the `Enum` and `Bounded` classes.

There are not a whole lot of exercises on this one, because there's a pretty limited scope of what you can do with these classes.

I included deriving-via and the `generic-data` package, in a "don't worry about how this works, but here's how you do it" presentation -- perhaps a bold choice, but it nicely fills in the gaps in GHC's stock deriving capability with little fuss, and I'm personally of the opinion that it's time to start talking about `via` as a normal thing to use.

### DIFF
--- a/en/lessons/typeclasses/enumeration.md
+++ b/en/lessons/typeclasses/enumeration.md
@@ -1,4 +1,560 @@
 ---
-version: "0.0.1"
+version: "1.0.0"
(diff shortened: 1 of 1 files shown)


Kleidukos (approved):

Kleidukos:
I agree with the normalisation of Deriving Via

PR #50, opened 2022-06-18
new lesson: Show and Read

chris-martin:
- I emphasized that the main purpose of `Show` is for ghci and that this class is not the end-all solution for building strings.
- I went into a bit of detail about `showsPrec` because there's a bunch you have to understand to be able to write well-behaved `Show` instances, and for opaque types you do sometimes want to do that. The section doubles as a rough introduction to Haskell precedence rules.
- I included only a passing mention of `Read`, because it's tremendously complicated and I have no idea what real use cases it has. I can work on it more if you like, but the cost/benefit here seems very poor.

### DIFF
--- a/en/lessons/typeclasses/string-representation.md
+++ b/en/lessons/typeclasses/string-representation.md
@@ -1,4 +1,681 @@
 ---
-version: "0.0.1"
+version: "1.0.0"
 title: "Representing values as strings: Show and Read"
 ---
+
+{% include toc.html %}
+
+This lesson uses the deriving-strategies language extension.
+
+```haskell
+{-# language DerivingStrategies #-}
+```
+
+A few imports and other language extensions are introduced as needed.
+
+## How GHCi prints results
+
+Some things you enter into GHCi don't result in anything being printed in response. For example, imports:
+
+```console?lang=haskell&prompt=ghci>,ghci|
+ghci> import Data.IORef
+ghci> import Data.Foldable
+```
+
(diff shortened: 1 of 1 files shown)


en/lessons/typeclasses/string-representation.md
```diff
@@ -1,4 +1,673 @@
...
+### Stock deriving
+
+`Show` can be stock-derived for any enumeration type. We did this without mentioning it in the previous lesson.
+
+```haskell
+data Brightness = Light | Dark
+    deriving Show
+
+data Hue = Red | Green | Blue
+    deriving Show
```
Kleidukos: It's nice that you expose the reader to "stock derivation". Could you perhaps include the `stock` deriving keyword as well? This is valid for the other instances of `deriving Show`

Kleidukos (changes requested):
Thank you very much for this PR. I only have one remark regarding the usage of deriving strategies (which is good to showcase!).

en/lessons/typeclasses/string-representation.md
chris-martin: Yeah, I'm a big fan of the explicit deriving strategies. I think I've discussed "stock derivation" in every lesson so far. Should we just go ahead and do it everywhere, and add a note style guide that deriving strategies should always be used?

Kleidukos: yes!

chris-martin: Cool, done.

Kleidukos (approved):
:+1:

PR #54, opened 2022-07-18
Foldable and Functor

chris-martin:
I've gotten through `Foldable` and `Functor`, and I think it's best to cut this lesson off there - this one doesn't need to get any longer than it already is, and `Traversable` will really work better in a new lesson after `Applicative` and `Monad`. I had originally thought I could get through an introduction to traversal by just always using `IO` as the context and hand-waving over the `Applicative` constraint, but the more I think of it the more I think we need to get to `Applicative`/`Monad` first to solidify the "action" concept and to get to expand the set of tools we can use in example code.

This lesson is now retitled to "Constructor class basics: Foldable and Functor" because that's really the common theme here, how to think about what a constructor class is. `Functor` is the classical introduction to constructor classes, but it's a little bit overly mind-bending due to the number of type parameters involved as we shift from one container type to another, and because it requires generalizing over two different conceptual metaphors, containers and actions. Starting with `Foldable` (and sidestepping most of its complexity) I think offers a more gentle ramp appropriate to the fairly quick pace of haskell school. It's easier to get a grasp on what kinds of things are `Foldable` because this class pretty much only describes container data structures.

### DIFF
--- a/_data/contents.yml
+++ b/_data/contents.yml
@@ -16,6 +16,7 @@ typeclasses:
   - enumeration
   - string-representation
   - monoids
-  - iteration
+  - constructor-classes
   - functors
+  - traversal
   - numbers
--- a/en/lessons/typeclasses/constructor-classes.md
+++ b/en/lessons/typeclasses/constructor-classes.md
@@ -1,0 +1,970 @@
+---
+version: "1.0.0"
+title: "Constructor class basics: Foldable and Functor"
+---
+
+{% include toc.html %}
+
+This lesson covers topics from these modules:
+
+```haskell
+import Data.Foldable
+import Data.Functor
+```
+
+We will discuss the following functions:
+
+|----------------------------------------------------------------------------------------------|------------------------------------------------------------------|
+| Function                                                                                     | Type                                                             |
+|----------------------------------------------------------------------------------------------|------------------------------------------------------------------|
+| [`toList`](https://hackage.haskell.org/package/base/docs/Data-Foldable.html#v:toList)        | `(Foldable t) => t a -> [a]`                                     |
+| [`foldr`](https://hackage.haskell.org/package/base/docs/Data-Foldable.html#v:foldr)          | `(Foldable t) => (a -> b -> b) -> b -> t a -> b`                 |
+| [`foldMap`](https://hackage.haskell.org/package/base/docs/Data-Foldable.html#v:foldMap)      | `(Foldable t, Monoid b) => (a -> b) -> t a -> b`                 |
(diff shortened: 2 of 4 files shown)


_data/contents.yml
```diff
@@ -16,6 +16,7 @@ typeclasses:
   - enumeration
   - string-representation
   - monoids
-  - iteration
+  - constructor-classes
   - functors
```
Kleidukos, line 20: It would seem that this lesson is absorbed by the current one, is that right?

why-not-try-calmer:
@chris-martin Just for my understanding, how does the part about `Foldable` relate to what you had written [here](https://github.com/haskellfoundation/HaskellSchool/pull/52#issuecomment-1170432124)? 

As far as I can see the ground covered here goes a long way toward a lesson about folds. Sure, I could imagine additional contents tying up some loose ends about practical use-cases (i.e. early termination/recursion with `foldr`, common patterns by combining with Applicatives, etc) but I am not sure whether I should suggest additions to this chapter or submit a short addendum (admittedly that would be a little bit weird for the current format).

What do you think?

_data/contents.yml
chris-martin: Yeah, that's right. I had previously used "iteration" as the tag when the lesson was going to lead up to Traversable, since the whole thing was going to build up to what has been described as ["essence of the iterator pattern"](https://www.cs.ox.ac.uk/jeremy.gibbons/publications/iterator.pdf), but since dropping Traversable from the lesson I renamed it from "iteration" to "constructor-classes".

chris-martin:
@why-not-try-calmer The biggest question we don't answer in this lesson, because I think it's too difficult and too detached from the larger aim of covering the core typeclasses, is how you should decide whether to use a left/right or strict/nonstrict fold. We also don't at all discuss how `toList` and `foldr` are actually related to one another: We present "folding" as a way to "reduce" a list to a summary value, and yet folding with `(:)` *builds up* a new structure comparable to the orginal that is not actually "reduced" at all - leading into this idea that I guess is what we call "co-recursion".

What I was envisioning was a separate lesson dedicated to non-strict evaluation, that would cover a lot of the same stuff that the `Data.Foldable` module documentation currently discusses, but perhaps divorced from the `Foldable` class. I don't really know where it fits into Haskell School, though, because I suppose the entire school is not planned yet; I think the two lesson groups that we presently have outlined are too beginner-level for it.

_data/contents.yml
Kleidukos: I realise now that the GitHub UI isn't the best when I thought I had highlighted one particular line. The `functor` lesson is also absorbed, right?

chris-martin: Oh, I see. The one called "functors" is Applicative/Monad.

